PR #16, opened 2023-04-13
various fixes to generics

nikomatsakis:
These are now sufficient to autogenerate `List` and `ArrayList` (albeit removing some troublesome overloaded methods etc).


nikomatsakis:
No idea why CI/CD is not triggering

PR #19, opened 2023-04-16
Add throw/catch support, including a crate error type and up/downcasting along the way

chase-ok:
The goal is to add support for catching exceptions of a certain type thrown by Java code. This requires:
* a way to check for and model thrown objects as an `Err` case,
* a way to check if a Java object is an instance of a particular class and downcast the object if it is

Most `jni` checked functions already return a `JavaException` error that is just a flag indicating an exception occurred. This commit introduces new crate `Error` and `Result` types that instead contain an actual reference to the thrown object. This can be a local ref inside of `Jvm::with` or a global ref when `with` returns.

Instance of checks and downcasting generally require a pointer to the target Java class object, so we add a `fn class()` to the `JavaObject` trait. For most cases, this will still be cached in a static `OnceCell` and replaces the existing `cached_class()` fn.

I also noticed the `mod java {}` was getting a bit large so I've moved it over to a separate file. I assume long-term this will be replaced by a single declaration instead of relying on copied `javap` outputs.

---

The end result is an extra call in the logger example that we can catch on:
```rust
l.throw_something()
    .catch::<Throwable, _>(|t| t.print_stack_trace())
    .execute(jvm)?;
```


macro/src/codegen.rs
```diff
@@ -267,20 +269,23 @@ impl SpannedClassInfo {
...
                             jvm: &mut Jvm<'jvm>,
                             (): (),
-                        ) -> duchess::Result<Self::Output<'jvm>> {
+                        ) -> duchess::Result<'jvm, Self::Output<'jvm>> {
                             #(#prepare_inputs)*
 
-                            let class = cached_class(jvm)?;
+                            let class = <#ty>::class(jvm)?;
+                            // XX: safety
+                            let jclass = unsafe { JClass::from_raw(class.as_jobject().as_raw()) };
```
chase-ok, line 277: Realized this after the fact, but there's a built-in `Into` that can do this without an unsafe. Will fix as part of the next PR

src/cast.rs
```diff
@@ -0,0 +1,166 @@
...
+        jvm: &mut crate::Jvm<'jvm>,
+        input: J::Input<'jvm>,
+    ) -> crate::Result<'jvm, Self::Output<'jvm>> {
+        let instance = self.op.execute_with(jvm, input)?;
+        let jobject = instance.as_ref().as_jobject();
+
+        let class = To::class(jvm)?;
+        let jclass = class.as_jobject();
+        // Safety: XX class is a Local<Class> which can only point to Java class objects
+        let jclass = unsafe { std::mem::transmute::<Jail<JObject>, Jail<JClass>>(jclass) };
```
chase-ok, line 80: ditto

src/cast.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::{JavaObject, JvmOp, Local, jvm::JavaObjectExt, Jvm};
+
+/// A trait to represent safe upcast operations for a [`JavaObject`].
+///
+/// # Safety
+///
+/// Inherits the rules of [`JavaObject`], but also `S` must be a valid superclass or implemented interface of `Self`.
+/// XX: would this actually allow unsafe behavior in a JNI call? or is it already checked/enforced?
+///
+/// XX: having to impl Upcast<T> for T on each struct is pretty annoying to get AsRef<T> to work without conflicts
```
nikomatsakis, line 17: these are auto-generated, right?

macro/src/codegen.rs
```diff
@@ -124,7 +125,20 @@ impl SpannedClassInfo {
...
                 where
                     #(#java_class_generics: JavaObject,)*
-                {}
+                {
+                    fn class<'jvm>(jvm: &mut Jvm<'jvm>) -> duchess::Result<'jvm, &'static Global<java::lang::Class>> {
+                        #cached_class
+                    }
+                }
+
+                unsafe impl<#(#java_class_generics,)*> ArrayElement for #struct_name<#(#java_class_generics,)*>
```
nikomatsakis: I feel like I changed how this works...merge conflict?

src/error.rs
```diff
@@ -5,14 +5,21 @@ use thiserror::Error;
 
 use crate::{java::lang::Throwable, Global, Jvm, Local};
 
+/// Result returned by most Java operations that may contain a local reference
+/// to a thrown exception.
 pub type Result<'jvm, T> = result::Result<T, Error<Local<'jvm, Throwable>>>;
+
+/// Result returned by [`crate::Jvm::with()`] that will store any uncaught
+/// exception as a global reference.
```
nikomatsakis, line 13: Question: If exceptions are "exceptional", is it worth the optimization of having local references? Would be simpler to just *always* convert the exception to a global reference, no?

src/catch.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use jni::{
+    objects::{AutoLocal, JObject},
+    JNIEnv,
+};
+
+use crate::{cast::Upcast, error::Error, java::lang::Throwable, JvmOp, Local};
+
+/// Plumbing utility to check the exception state of the JVM thread convert it into a [`crate::Result`].
+// XX: many of the jni crate checked methods do this automatically. We only need this if/when we invoke
```
nikomatsakis, line 11: I'm confused -- those methods don't, presumably, convert to `Error::Thrown`, right, but rather to some JNI-internal result variant?

src/catch.rs
```diff
@@ -0,0 +1,92 @@
...
+// unchecked methods.
+pub fn try_catch<'jvm>(env: &mut JNIEnv<'jvm>) -> crate::Result<'jvm, ()> {
```
nikomatsakis, line 13: I don't see any calls to this function?

nikomatsakis (changes requested):
I'm a bit confused about a few aspects of this one

src/catch.rs
chase-ok: I think we can delete this fn, but for posterity, the `jni` crate methods automatically check for an exception flag and then return a bare `jni::errors::Error::JavaException` without storing a reference to the exception object.

The bigger issue I was worried about was proceeding to make further Java calls without first checking and clearing that flag. It would cause the exception to be rethrown in an odd place. But, it looks like the `jni` crate has put this check everywhere.

src/catch.rs
chase-ok: Yep, I had written it assuming the `*_unchecked` version of JNI functions wouldn't check for the exception flag and we'd use this in the next PR that moves to `*_unchecked`, but that was wrong. It looks like it's checked everywhere.

Will delete this!

src/error.rs
chase-ok: It's a good question. In my head, Java exceptions are as exceptional as `Result` is in Rust---not very exceptional. Folks use them to express expected conditions all the time and they're fairly optimized within the JVM. For example, the auth code needs to check for the expected case of a few different auth failure types. They're more rare than the happy case, but not infrequent. 

Loosely speaking, I think it would be like declaring a single `Result<T, Box<dyn Error + Send>>` alias in Rust and using it everywhere. It would avoid some typing and performs fine most of the time. 

---

Thinking more about it, we only materialize the exception object during `.catch()` or during `.execute()`. We could keep it as a local for `.catch()` and then leave the result type as a global.

I'm assuming you're asking from the perspective of simplifying the result type to remove that `'jvm` lifetime?

macro/src/codegen.rs
chase-ok: I think this was on an older commit I didn't squash after I rebased. Were you looking at the final diff?

src/error.rs
chase-ok: I think I've convinced myself the Global cost is probably worth the simplicity for most cases.

src/catch.rs
chase-ok: Coming back, I think we actually need this after our `jni` crate calls. A user that inspects the error case of `.execute()`  inside of `Jvm::with` won't get the materialized thrown object. We can't do the materialization by default in the `From` impl either, because it needs a `&mut Jvm` to get the reference.

nikomatsakis:
Can you cleanup the history on this one? e.g. merge with origin/main and then do a "git reset --soft origin/main"? right now it's got a lot of drift. We may want to merge #25 first too

PR #20, opened 2023-04-16
Cache MethodID for constructor & method calls

chase-ok:
Use the same `OnceCell` caching mechanism for looking up method IDs for constructor and method calls. This forces us to switch to the `*_unchecked()` JNI functions. The checks in the regular versions should be redundant with our own type system. It also avoids parsing the method signature string on each call!

### DIFF
--- a/macro/src/codegen.rs
+++ b/macro/src/codegen.rs
@@ -461,6 +461,9 @@ impl SpannedClassInfo {
 
                     #(#prepare_inputs)*
 
+                    // Cache the method id for this method -- note that we only have one cache
+                    // no matter how many generic monomorphizations there are. This makes sense
+                    // given Java's erased-based generics system.
                     static METHOD: OnceCell<JMethodID> = OnceCell::new();
                     let method = METHOD.get_or_try_init::<_, duchess::Error<Local<java::lang::Throwable>>>(|| {
                         let class = <#this_ty>::class(jvm)?;
--- a/macro/src/codegen.rs
+++ b/macro/src/codegen.rs
@@ -278,6 +278,9 @@ impl SpannedClassInfo {
 
                             let env = jvm.to_env();
 
+                            // Cache the method id for the constructor -- note that we only have one cache
+                            // no matter how many generic monomorphizations there are. This makes sense
+                            // given Java's erased-based generics system.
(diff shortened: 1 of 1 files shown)


macro/src/codegen.rs
```diff
@@ -458,11 +461,22 @@ impl SpannedClassInfo {
 
                     #(#prepare_inputs)*
 
+                    static METHOD: OnceCell<JMethodID> = OnceCell::new();
```
nikomatsakis, line 464:
```suggestion
                    // Cache the method id for this method -- note that we only have one cache
                    // no matter how many generic monomorphizations there are. This makes sense
                    // given Java's erased-based generics system.
                    static METHOD: OnceCell<JMethodID> = OnceCell::new();
```

macro/src/codegen.rs
```diff
@@ -273,19 +273,21 @@ impl SpannedClassInfo {
...
 
-                            let o = env.new_object(
-                                jclass,
-                                #descriptor,
-                                &[
-                                    #(JValue::from(#input_names),)*
-                                ]
-                            )?;
-                            // env.new_object() internally calls check_exception!()
+                            static CONSTRUCTOR: OnceCell<JMethodID> = OnceCell::new();
```
nikomatsakis, line 281:
```suggestion
                            // Cache the method id for the constructor -- note that we only have one cache
                            // no matter how many generic monomorphizations there are. This makes sense
                            // given Java's erased-based generics system.
                            static CONSTRUCTOR: OnceCell<JMethodID> = OnceCell::new();
```

macro/src/codegen.rs
```diff
@@ -621,6 +635,7 @@ impl Signature {
     /// Returns an appropriate `impl type` for a funtion that
     /// returns `ty`. Assumes objects are nullable.
     fn output_type(&mut self, ty: &Option<Type>) -> Result<TokenStream, SpanError> {
+        // XX: do we need the non_repeating transform here? Shouldn't be allowed in return position
```
nikomatsakis, line 638: True, we could avoid it and just add the extra match arm with a panic

nikomatsakis (changes requested):

nikomatsakis (approved):

PR #21, opened 2023-04-17
Add by_type! macro, multiple catch arms, finally block

chase-ok:
The `by_type!` macro allows users to write a pseudo-match expression
that tries to downcast a Java object into a given class or interface.

It's intended mainly for inspecting the output Java APIs that can throw different
kinds of exceptions for different error cases. 

---

Also add support for multiple catch arms and a `finally` block.

### DIFF
--- a/macro/src/codegen.rs
+++ b/macro/src/codegen.rs
@@ -6,6 +6,7 @@ use crate::{
     span_error::SpanError,
 };
 use inflector::Inflector;
+use once_cell::sync::OnceCell;
 use proc_macro2::{Ident, Literal, Span, TokenStream};
 use quote::quote_spanned;
 use rust_format::Formatter;
@@ -172,10 +173,8 @@ impl SpannedClassInfo {
     fn upcast_impls(&self) -> TokenStream {
         let struct_name = self.struct_name();
         let java_class_generics = self.class_generic_names();
-        self.info
-            .extends
-            .iter()
-            .chain(&self.info.implements)
+        self
+            .resolve_upcasts()
             .map(|r| {
                 let mut sig = Signature::new(&Id::from("supertrait"), self.span, &self.info.generics);
                 let tokens = sig.forbid_capture(|sig| sig.class_ref_ty(r)).unwrap();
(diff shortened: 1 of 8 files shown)


macro/src/codegen.rs
```diff
@@ -189,6 +188,22 @@ impl SpannedClassInfo {
             .collect()
     }
 
+    // XX: Clearly, we'll need more sophisticated resolution of what types we descend from, but for now we can at least
+    // inject the "everything is an Object" root.
```
nikomatsakis, line 192: I wouldn't expect it to be necessary for us to inject Object -- but I guess you are just leaving a "TODO" to implement the transitive closure?

src/cast.rs
```diff
@@ -164,3 +159,71 @@ where
         }
     }
 }
+
+/// Branch on the instance type of a Java object. It will execute (and only execute) the first match arm that has a type
```
nikomatsakis, line 163: Hmm, I feel like, rather than going the macro route, we could instead do something with methods. I saw your comment about the subtle difference of chained `catch` methods, but that assumes we do a naive chaining.

We could instead do something like this...

```rust
jvm_op
    .catch_block()
    .catch(|x: Bar| ...)
```

where `catch_block` is a method on `JniOp` and `catch(self)` is an (inherent) method on the `Catch` operator type itself. The idea would be that `catch_next` doesn't apply to exceptions that occur inside of `catch_first`, and instead adds arms into the catch block. We probably want a  `catch_otherwise` or `finally` as well. We could have a complete method to exit the catch block if we wanted, or simply have it also implement `JniOp`.

Another option might be...

```rust
x.catch_first(....).catch_next(...).finally(...)
```

you could even do `x.catch().catch()` and rely on the Rust method handle (which gives precedence to inherent methods), but that seems a bit subtle and error prone for my taste.

nikomatsakis (changes requested):

macro/src/codegen.rs
chase-ok: Yeah, meant as a TODO for the full transitive closure impl. I just needed the `Object` root everywhere to make the "inspect a Java Object" example pass

src/cast.rs
chase-ok: You're right, I think I'm conflating a couple of issues here.

We can fix the chained catches with something like `catch_block()` or `catch_one_of()`. I'l go ahead and impl that since it's cleaner than the macro for that case. 

That said, I do expect folks to need to inspect and branch on the type of a returned object, be it an exception or regular data. In the auth example, I imagine the real code looking like

```rust
let auth_result: Result<User, AuthFailure> = Jvm::with(|jvm| {
    let http_request = HttpRequest::new(/* ... */);
    match authenticator.authenticate(http_request).execute(jvm) {
        Ok(authenticated_user) => Ok(Ok(build_user(authenticated_user))),
        Err(duchess::Error::Thrown(t)) => by_type! {
            with jvm match t => {
                InvalidSignatureException as t => Ok(Err(AuthFailure::InvalidSignature(t.get_message().execute(jvm)?))),
                InternalError => Ok(Err(AuthFailure::InternalError)),
                // ...
            }
        }
        Err(e) => Err(e),
}).unwrap();
```

I don't think trying to put that "into Rust" logic in the catch JvmOp will work since they're not types that can be put into Java land.

chase-ok: Added the `catching()` fn in the next commit

nikomatsakis: Hmm, let's talk this one out.

nikomatsakis (approved):
OK, let's land this, I'm not sure if it's in the final state we want, but it's getting there.

nikomatsakis:
Is this PR dependent on #19 ?

nikomatsakis:
Looks like not.

PR #23, opened 2023-04-17
Add test crate

fpoli:
This PR makes `duchess` read the `CLASSPATH` variable. Since the programs in `examples/` cannot set custom environment variables, I added a `test-crates` workspace where it's actually possible to write build scripts and test the `CLASSPATH`. Having a build script also makes it easy to add tests of libraries whose JAR is specified by URL, instead of being forced to add lots of bytecode to the repository. What do you think?

The only test crate that I added uses the JAR of a stable [Viper](https://github.com/viperproject) release because I care about that use case and because that JAR is full of useful edge cases to test the parser (e.g. `scala/$eq$colon$eq.class`).


chase-ok (approved):

macro/src/reflect.rs
```diff
@@ -63,11 +63,17 @@ impl JavaClass {
     fn parse_javap(&self, package_name: &JavaPath) -> Result<SpannedClassInfo, SpanError> {
         let mut command = Command::new("javap");
 
+        let classpath = match env::var("CLASSPATH") {
+            Ok(val) => val,
+            Err(e) => panic!("duchess cannot read the CLASSPATH environment variable: {e}"),
```
nikomatsakis, line 68: Panicking if we can't find CLASSPATH sems a bit extreme, no? Can we just not pass a `-cp` argument in that case?

macro/src/reflect.rs
```diff
@@ -63,11 +63,17 @@ impl JavaClass {
...
+        };
+
         command
+            .arg("-cp")
+            .arg(classpath)
```
nikomatsakis, line 73: This makes sense to me, but doesn't javap get the environment variables from the parent process anyway?

test-crates/viper/build.rs
```diff
@@ -0,0 +1,20 @@
...
+    let class_path = out_dir.join("classes");
+
+    let jar_data = ureq::get(JAR_URL).call().unwrap();
+    let mut jar_file = fs::File::create(jar_path.clone()).unwrap();
+    io::copy(&mut jar_data.into_reader(), &mut jar_file).unwrap();
+    zip_extract(&jar_path, &class_path).unwrap();
+
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed={}", class_path.display());
+    println!("cargo:rustc-env=CLASSPATH={}", class_path.display());
```
nikomatsakis, line 19: Hmm, if this pattern is useful, maybe we should provide some helper methods for it in duchess?

nikomatsakis (approved):

macro/src/reflect.rs
nikomatsakis: Oh right, I forgot, it doesn't seem to

nikomatsakis (approved):

PR #25, opened 2023-04-21
implement method listing; use main macro for JDK classes

### LINKED ISSUE #4: selective method mirroring
Right now we only support mirroring *all* the methods in a class:

```rust
duchess::java_package! {
    package foo;
    class Bar { * }
}
```

but we should support selective mirroring. What's a bit unclear is *how* to do that. By name? By full signature (to account for overloaded methods)? I'm inclined to just go with "by name" for simplicity, and because I imagine we'll want to support it for ergonomics anyway, maybe something like this?

```rust
duchess::java_package! {
    package foo;
    class Bar { Bar(), toString(), computeName() }
}
```

where `Bar()` refers to the constructor(s).

That would give us space to adopt full signatures in the future if that ever seemed necessary.

### LINKED ISSUE #12: Support parsing inner classes
Java inner class names are written as `OuterClass$InnerClass`. It looks like our javap parser is failing on the `$` token in type names:

```
error: invalid token `... .invoke.TypeDescriptor <<< $ >>> OfField<java.lang.Clas ...` at offset 1
   --> src/java/lang.rs:125:1
    |
```

### LINKED ISSUE #13: Support parsing static methods
It looks like our javap parser is failing after an unexpected `static` token before the return type for static methods:

```
error: unexpected token `...         public static  <<< int >>>  test();
                   d ...` at offset 1302, expected one of `["\"(\"", "\")\"", "\",\"", "\".\"", "\";\"", "\"<\"", "\">\"", "\"[\"", "\"extends\"", "\"implements\"", "\"{\"", "r#\"[a-zA-Z][a-zA-Z0-9]*\"#"]`
```

nikomatsakis:
This branch includes a number of bug fixes that let you use the macro to select which constructors/methods to translate. You now get errors if the list of classes is not complete or consistent. Use this to import things from JDK instead of writing code by hand or copy/pasting the output from javap. So we now have things like...

```rust
duchess! {
    ...
    
    package java.util;

    class List {
        add(E);
    }

    class ArrayList {
        ArrayList();
    }

    class Map {
        put(K, V);
    }

    class HashMap {
        HashMap();
    }

    ...
}
```

...obviously lots of small things to fix here, but overall seems good.

Fixes #4
Fixes #12 
Fixes #13


nikomatsakis:
I'm going to try and get into good merge discipline -- so requesting a review from @chase-ok :)

nikomatsakis:
I'm going to close this PR and break it up into smaller pieces.

PR #29, opened 2023-04-23
Fix thrown exceptions being buried during execute() calls

chase-ok:
A Duchess user can execute most `crate::JvmOp`s at any time inside of a `crate::Jvm::with()` call. The returned `Result` should correctly contain an `Error::Thrown` if a Java exception was thrown. If we waited to materialize errors until exiting `with` or in a `catching()` block, the execute call would return an internal `Error::Jni` instead and the exception would be buried.

`Error::Jni` should be reserved for unexpected JNI failures and not exposed to Duchess users for normal scenarios like a thrown exception.


nikomatsakis (changes requested):
I agree with the intent, but I don't see any obvious tests in this PR. Can we create a test that shows the change in behavior? Any thoughts on how we would do that?

Also, I have a question: is the idea that whenever a function returns a raw `jni::Result` value, that we should use `with_java_env` to convert it to a `duchess::Result` value, and it will ensure that we handle this case correctly? 

Either way, can you extend the mdbook (in the `book` directory) with an implementation page talking about how we handle thrown exceptions and the right conventions to follow when invoking jni methods? I'm imagining a page in the https://duchess-rs.github.io/duchess/internals.html section, probably a subpage, that is targeting contributors, not user-facing documentation.

src/error.rs
```diff
@@ -85,14 +62,110 @@ pub(crate) enum JniErrorInternal {
...
+                Err::<(), _>(jni::errors::Error::TryLock)
+            })
+            .unwrap_err();
+            assert!(matches!(err, Error::Jni(_)));
+            Ok(())
+        })
+        .unwrap();
+    }
+
+    // XX this should likely move to an integration test suite
```
chase-ok, line 157: Let me know if this suffices for now or if we want to tackle a more proper test suite first.

nikomatsakis (approved):
Great docs, thanks. For tests, seems good, I was thinking it'd be good to check various edge cases and check for thrown exceptions, but I think this will do. We should start tracking the various things to test.

PR #30, opened 2023-04-24
Flesh out auth example

chase-ok:
Create a more comprehensive authenticate and authorize example to help iterate on crate features and pain points.


examples/auth.rs
```diff
@@ -1,45 +1,279 @@
...
+                            AuthenticateError::Unathenticated(message)
+                        // XX: should we add a .is_instance() alias for try_downcast().is_ok()?
+                        } else if exception.try_downcast::<Throwable, java_auth::AuthenticationExceptionInvalidSecurityToken>().execute(jvm)?.is_ok() {
+                            AuthenticateError::InvalidSecurityToken
+                        } else if exception.try_downcast::<Throwable, java_auth::AuthenticationExceptionInvalidSignature>().execute(jvm)?.is_ok() {
+                            AuthenticateError::InvalidSignature
+                        } else {
+                            let message = exception.get_message().assert_not_null().into_rust(jvm)?;
+                            AuthenticateError::InternalError(message)
+                        }
```
chase-ok, line 110: Wild idea: could we add a macro'd enum of all downcastable types listed in the `java_package!` invocation? Something like

```rust
duchess::java_package! {
    package auth;
    
    // ...
    #[downcasting]
    class AuthenticationException { * }
    class Unauthenticated { * }
    class InvalidSecurityToken { * }
    class InvalidSignature { * }
}
```

which would then let this code look like

```rust
match exception.downcasting(jvm)? {
    AuthenticationExceptionDowncast::Unauthenticated(x) => { .. }
    AuthenticationExceptionDowncast::InvalidSecurityToken(_x) => { .. }
    AuthenticationExceptionDowncast::Else(x) => { .. }
}
```

nikomatsakis (commented):
This is awesome! It's really helpful to have this example. This looks a bit different than I expected, so I'm going to do a bit of tinkering and post up a revised proposal.

examples/auth.rs
nikomatsakis: That's a neat idea. I think we could do *something* like that yes. I like the idea of trying to make it possible to make authoring the Java code a bit more "rust-y".

examples/auth.rs
```diff
@@ -1,45 +1,279 @@
...
-    class AuthenticateResult { * }
-    class HttpAuth { * }
-    class HttpRequest { * }
+#[derive(Debug)]
+pub struct HttpRequest {
+    pub verb: String,
+    pub path: String,
+    pub body_hash: Vec<u8>,
+    pub params: HashMap<String, Vec<String>>,
+    pub headers: HashMap<String, Vec<String>>,
```
nikomatsakis, line 51: I don't love that we are authoring this struct by hand. I wonder if we can add some kind of option to have duchess auto-generate it -- or just do it always for record types. Something to think about.

examples/auth.rs
```diff
@@ -1,45 +1,279 @@
...
+        let auth = Jvm::with(|jvm| {
+            let auth = java_auth::HttpAuth::new().execute(jvm)?;
+            Ok(jvm.global(&*auth))
+        })?;
+        Ok(Self(auth))
+    }
+
+    pub fn authenticate(&self, request: &HttpRequest) -> Result<Authenticated, AuthenticateError> {
+        Jvm::with(|jvm| {
+            let request = request.into_java(jvm)?;
```
nikomatsakis, line 92: It'd be cool to autogenerate this too, I think, and maybe base it on serialization? Have to think about it.

examples/auth.rs
```diff
@@ -1,45 +1,279 @@
...
+impl HttpAuth {
+    pub fn new() -> duchess::GlobalResult<Self> {
+        let auth = Jvm::with(|jvm| {
+            let auth = java_auth::HttpAuth::new().execute(jvm)?;
+            Ok(jvm.global(&*auth))
+        })?;
+        Ok(Self(auth))
+    }
+
+    pub fn authenticate(&self, request: &HttpRequest) -> Result<Authenticated, AuthenticateError> {
```
nikomatsakis, line 90: Ultimately, I would like this to be one big `execute` call.

nikomatsakis (commented):
Some random notes -- mostly to myself -- I didn't have time to go into more detail.

Issue #6, opened 2023-04-10
jvm builder


nikomatsakis:
The [`Jvm`](https://github.com/nikomatsakis/duchess/blob/0a1dd1ec20bbe8aee52077ef149f6019a99a5108/src/jvm.rs#L63-L66) type is kind of a placeholder. The [docs](https://github.com/nikomatsakis/duchess/blob/main/book/src/jvm.md) describe how it should look eventually, e.g., with a builder that lets you set the classpath and other important options.

nikomatsakis:
See also @fpoli's comments here https://github.com/duchess-rs/duchess/pull/45#discussion_r1180341027

fpoli:
(I rephrase here my comments.)

Reading the ["Starting multiple JVMs"](https://duchess-rs.github.io/duchess/jvm.html#starting-multiple-jvms) section of the current documentation I got the (wrong) impression that sequential `Jvm::with(|jvm|..)`  calls are meant in the long term to start and stop multiple JVMs. However, via JNI one can only start at most one JVM for the full the *lifetime* of a process ([source](https://bugs.openjdk.org/browse/JDK-4712793)). That is, trying to start a second JVM after destroing one will always fail.

Re-reading the documentation with a fresh mind, I now see that the "multiple JVMs" of the documentation actually refers to multiple instances of the `Jvm` struct. PR https://github.com/duchess-rs/duchess/pull/47 should avoid any further confusion.

My suggestion of [`std::sync::OnceLock`](https://doc.rust-lang.org/nightly/std/sync/struct.OnceLock.html) is still valid; it's a concurrent `OnceCell`. That type makes it possible to initialize a global JVM with user-provided arguments.

PR #45, opened 2023-04-28
Test construction of a Scala class

fpoli:
This PR contains a few improvements (see the commit messages) and a test that runs the JVM with a custom `CLASSPATH`.


src/jvm.rs
```diff
@@ -105,12 +106,13 @@ pub trait IsVoid: Default {}
...
-        .build()
-        .unwrap();
+        .option("-Xcheck:jni");
+    if let Ok(classpath) = env::var("CLASSPATH") {
+        jvm_builder = jvm_builder.option(format!("-Djava.class.path={classpath}"));
+    }
+    let jvm_args = jvm_builder.build().unwrap();
 
     JavaVM::new(jvm_args).unwrap()
 });
```
fpoli, line 118: I just noticed that this is not the nice `Jvm::builder().add_classpath(..).launch(|jvm| ...)` described in the documentation.

fpoli: I find the builder pattern to be a bit strange, because it's not possible to build more than one JVM per process. What about changing the type of `GLOBAL_JVM` to be `OnceLock<JavaVM>`, so that the following is possible?
```rust
Jvm::builder().add_classpath(..).launch()?;
Jvm::with(|jvm| ...)?;
Jvm::with(|jvm| ...)?;
Jvm::with(|jvm| ...)?;
```
Calling `Jvm::builder().launch()` on an already initialized JVM would return an error, as well as calling `Jvm::with(..)` on a non-initialized JVM.

nikomatsakis: I don't know what a "once lock" is, but that is kind of what I had in mind -- except that I figured we would have a default so if you called `Jvm::with` without having invoked `launch`, you got a default configuration.

nikomatsakis: The one constraint I want to maintain is that you have at most one `Jvm` handle available at a time -- i.e., nested `Jvm::with` should not work. I'm not sure how important this is, but it lets us do nested frames if we ever want that.

nikomatsakis (approved):

src/jvm.rs
nikomatsakis: let's follow up on https://github.com/duchess-rs/duchess/issues/6

PR #47, opened 2023-05-02
Clarify behavior of sequantial `Jvm::with` calls

fpoli:
If correct, this edit clarifies my confusion in https://github.com/duchess-rs/duchess/pull/45#discussion_r1180542126.


nikomatsakis (approved):
Yes, this matches my expectations.

PR #48, opened 2023-05-02
Implement `JvmBuilder`

### LINKED ISSUE #6: jvm builder

fpoli:
Implementation of what described in https://github.com/duchess-rs/duchess/issues/6. The global JVM is stored in a `once_cell::sync::OnceCell` to ensure that it's only constructed once via duchess' API. Other types that I considered and discarded:
* `once_cell::sync::Lazy` has no way of postponing the declaration of the constructor,
* `once_cell::race::onceBox` doesn't prevent concurrent invocations of the constructor, 
* `std::sync::OnceLock` depends on unstable features.

Fixes #6.

### DIFF
--- a/src/jvm.rs
+++ b/src/jvm.rs
@@ -9,7 +9,8 @@ use crate::{
 };
 
 use std::{
-    env,
+    borrow::Cow,
+    fmt::Display,
     marker::PhantomData,
     ops::{Deref, DerefMut},
     ptr::NonNull,
@@ -19,7 +20,7 @@ use jni::{
     objects::{AutoLocal, GlobalRef, JObject, JValueOwned},
     sys, InitArgsBuilder, JNIEnv, JavaVM,
 };
-use once_cell::sync::{Lazy, OnceCell};
+use once_cell::sync::OnceCell;
 
 /// A "jdk op" is a suspended operation that, when executed, will run
 /// on the jvm, producing a value of type `Output`. These ops typically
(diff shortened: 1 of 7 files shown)


chase-ok (commented):
Just a heads up, I’m planning to remove the `jni` dependency and use the raw `jni_sys`. I think this interface change is compatible but there will some minor changes to the impl and error representation!

src/jvm.rs
```diff
@@ -178,6 +180,48 @@ impl<'jvm> Jvm<'jvm> {
...
+    }
+
+    pub fn custom(mut self, opt_string: impl AsRef<str> + Into<Cow<'args, str>>) -> Self {
+        self.args_builder = self.args_builder.option(opt_string);
+        self
+    }
+
+    fn build_jvm(self) -> crate::GlobalResult<JavaVM> {
+        let jvm_args = self.args_builder.build()?;
+        let jvm = JavaVM::new(jvm_args)?;
```
chase-ok, line 209: Rather than fail here if a JVM already exists, what if we just grabbed a pointer to it instead? This would allow duchess to be used as part of a library where consumers can "bring their own JVM".

fpoli: Good point. It would be useful for the case where `build_jvm` is called from `Jvm::with(..)`. In `get_or_default_init_jvm` it should be possible to catch the error result and then call [`JNI_GetCreatedJavaVMs`](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#JNI_GetCreatedJavaVMs) to obtain the existing JVM. However, as far as I know `JNI_GetCreatedJavaVMs` is not exposed by `jni`. Instead of starting to use `jni_sys` in this PR, it might be better to wait until your planned refactoring.

When calling `build_jvm` from `Jvm::builder().set_classpath(..).launch(..)`, the failure is intended because the consumer explicitly asked for a JVM with a certain classpath. Does that sound good?

src/jvm.rs
```diff
@@ -178,6 +180,48 @@ impl<'jvm> Jvm<'jvm> {
...
+        Ok(jvm)
+    }
+
+    /// Launch a new JVM, returning an error if one already exists.
+    pub fn launch<R>(
```
nikomatsakis, line 214: I suspect most users would want to configure the JVM at the beginning but not actually "launch" it. Something like:

```rust
fn main() {
    Jvm::builder().foo.bar.build()?;
    ...
    do_stuff();
}

fn do_stuff() {
    Jvm::with(|jvm| { ... });
}
```

This could just be an alias for `launch(|_| Ok(()))`.

I *suspect* it would also be useful to have something `try_build` that doesn't consider it an error if the jvm is already launched. 

But we can do this in a later PR.

nikomatsakis (approved):

PR #49, opened 2023-05-05
Parse interfaces with multiple extends

fpoli:
Fix the grammar to support `interface Foo extends A, B, C`.


nikomatsakis (approved):
Oh, for interfaces? I forgot those use the extends keyword. OK.

PR #51, opened 2023-05-07
Simplify JvmOp, refine Logger example

nikomatsakis:
This branch makes some simplifications (to pave way for future improvements) and also implements a refined logger example. This refined logger example demonstrates a pattern we'd like to support.

Added:

* new logging example, including multiple tests demonstrating the ways it can be used
* support for static methods
* more types from JDK (e.g., `java.util.Date`)
* `global` combinator to create a global reference
* default JVM loads classpath (cc @fpoli)

Removed and simplified:

* the `inspect` combinator and the `Input` field from `JvmOp`
* the `catching` combinator
* the various traits (e.g., `IntoJava`, `JavaMethod`) are now strictly aliases of `JvmOp`


src/jvm.rs
```diff
@@ -116,15 +110,29 @@ pub struct Jvm<'jvm> {
...
     }
 
+    /// Creates a JVM builder with default options.
+    /// This is what you get if you invoke `Jvm::with` without launching a JVM.
+    ///
+    /// Default options currently include:
+    ///
+    /// * classpath derived from `CLASSPATH` in the environment.
+    pub fn default_builder<'args>() -> JvmBuilder<'args> {
+        JvmBuilder::new()
```
fpoli, line 130:
```suggestion
        JvmBuilder::default_options()
```

src/jvm.rs
```diff
@@ -116,15 +110,29 @@ pub struct Jvm<'jvm> {
 impl<'jvm> Jvm<'jvm> {
     fn get_or_default_init_jvm() -> crate::GlobalResult<&'static JavaVM> {
         GLOBAL_JVM.get_or_try_init(|| {
-            let jvm = JvmBuilder::new().build_jvm()?;
+            let jvm = JvmBuilder::default_options().build_jvm()?;
```
fpoli, line 113: This should be a be a bit more likely to stay in sync with the documentation of `default_builder`.
```suggestion
            let jvm = Jvm::default_builder().build_jvm()?;
```

nikomatsakis:
OK, I reached a point where I am approximately satisfied with the `auth` example. The intent is that all of the `ToRust` impls will be autogenerated with a derive. All those patterns seem achievable -- one interesting bit is the capturing of the original java object. I named the field `this` because it seemed like a good convention that the derive could recognize. (i.e., you can capture the results of various getters and, optionally, capture the original object). The idea is that deriving on an enum will be generate the downcast chain, and you will specify the base type and the types for the various variants.

nikomatsakis:
I still haven't updated the user docs. Will do that later this morning.

nikomatsakis:
Rebased and did more work on this in the meantime. It's getting *close* to what I think the final shape ought to be. Still ill-documented. Some very brief notes:

* The `catch::<E>` combinator converts the output type from `O` to `Result<O, E>`, where `E` is some java throwable.
* The `to_rust` combinator converts from a Java to a Rust type; if `O` is a `Result` (i.e., from `catch`), then it converts the `Ok` and `Err` parts. So, if you have a fallible Java Rust, and you have a `ToRust` impl for both the success parts and the Java exception `E` you want to catch, you can do `foo.catch::<E>().to_rust()` and get back a Rust result. (It *may* be that we want to make `catch` always do a `to_rust` operation, but I kind of liked it separate.)
* There is now a `execute` method that just wraps `Jvm::with`; this works great if your result type is global or Rust, which I think ought to be the recommended path.
* For Rust types, the expectation is that they will implement `JvmOp` for `&Self`, but also that they will implement `ToJava<T>` for the output type. This split exists to account for the collection types -- e.g., there is more than one Java type you might want to translate `Vec` into, and you can't always let subtyping solve it (consider `ArrayList<ArrayList<E>>` vs `ArrayList<List<E>>`). The same isn't true for user-defined types, I don't think. The intent is that the auto-derive will invoke `.to_java::<T>()` with the Java type `T` derived from the Java class, but when users construct calls, they can typically just use the Rust value in place (because it implements `JvmOp`). The only exception would be maps/collections, which would need to use `.to_java::<T>()` explicitly.

Things I am unsure of:

* Do we want to make *all* conversions from Rust to Java explicit? i.e., not implement `JvmOp` for Rust types? I tried it but found that it seemed annoying, and I hit some kind of ergonomic pain points involving auto-ref. The latter are probably solvable.
* It's annoying that you have to write `.to_java::<T>()` and can't rely on type inference, but I don't see an obvious fix for this. It arises because of the upcast impls and Rust not being, apparently, smart enough to figure out the right choices. I haven't played with this too much either. In principle it *seems* inferrable much of the time. OTOH, if users don't have to write these calls themselves most of the time... maybe it's fine.

PR #52, opened 2023-05-07
Use `jni-sys` directly instead of through `jni`

chase-ok:
The `jni` crate provides a thin wrapper on top of the `jni-sys` headers and we were often writing more complicated code to use it instead of making direct `jni-sys` calls. It also gives us more flexibility around exception handling, `libjvm` linking, and users brining their own JVM instance.

The `raw` mod now contains most of the newtype wrappers we need around `jni-sys` pointers.

This commit adds an optional `dylibjvm` feature that gives the same dynamic linking behavior as the `jni` crate. Without it, it expects `libjvm` to be statically linked. It also adds a friendly thrown-error message.

Some of the initialization and string-passing code should be marginally faster since we avoid some copies into and out of cesu8 and c encodings.

### DIFF
--- a/src/libjvm.rs
+++ b/src/libjvm.rs
@@ -1,0 +1,74 @@
+use crate::GlobalResult;
+
+/// Virtual table for top-level libjvm functions that create or get JVMs.
+#[allow(non_snake_case)]
+pub(crate) struct Libjvm {
+    pub JNI_CreateJavaVM: unsafe extern "system" fn(
+        pvm: *mut *mut jni_sys::JavaVM,
+        penv: *mut *mut std::ffi::c_void,
+        args: *mut std::ffi::c_void,
+    ) -> jni_sys::jint,
+    pub JNI_GetCreatedJavaVMs: unsafe extern "system" fn(
+        vmBuf: *mut *mut jni_sys::JavaVM,
+        bufLen: jni_sys::jsize,
+        nVMs: *mut jni_sys::jsize,
+    ) -> jni_sys::jint,
+}
+
+#[cfg(feature = "dylibjvm")]
+mod dynlib {
+    use std::path::{Path, PathBuf};
+
+    use libloading::Library;
+    use once_cell::sync::OnceCell;
+
+    use super::*;
+    use crate::Error;
+
+    static LIBJVM: OnceCell<Libjvm> = OnceCell::new();
+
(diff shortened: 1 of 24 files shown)


src/libjvm.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[allow(non_snake_case)]
+    fn load_libjvm_at(path: &Path) -> GlobalResult<Libjvm> {
+        (|| {
+            let lib = unsafe { Library::new(path) }?;
+            let JNI_CreateJavaVM = *unsafe { lib.get(b"JNI_CreateJavaVM\0") }?;
+            let JNI_GetCreatedJavaVMs = *unsafe { lib.get(b"JNI_GetCreatedJavaVMs\0") }?;
+            Ok(Libjvm {
+                JNI_CreateJavaVM,
+                JNI_GetCreatedJavaVMs,
+            })
```
fpoli, line 39: I investigated a bit the `SIGSEGV` in CI: it's because `Library` [unloads the library on drop](https://docs.rs/libloading/latest/libloading/struct.Library.html#method.close).

nikomatsakis (commented):
There is more code here than I expected! But I do think it makes sense, especially if it will let us be more efficient (e.g., by avoiding enum wrappers). I saw there's a segfault though?

src/ref_.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: from the guarantees of Local::from_raw, we know obj is a live local ref to an instance of T
+        unsafe { T::from_raw(self.obj) }
+    }
+}
+
+/// An owned global reference to a non-null Java object of type `T`. The reference will be freed when dropped.
+pub struct Global<T: JavaObject> {
+    jvm: JvmPtr,
```
nikomatsakis, line 73: given that there is only one jvm per process, do we really need to hold on to this pointer? can't we store it somewhere global?

nikomatsakis (approved):
Seems good to me. I think long term we are going to want to do this.

src/ref_.rs
chase-ok: Oops, yes, we definitely can!

chase-ok: Removed in next commit

fpoli (approved):

PR #53, opened 2023-05-10
Fix race condition in JVM creation

chase-ok:
Fixes a race condition when multiple threads try to create or look up existing JVMs at the same time by sequencing the calls behind the `GLOBAL_JVM` sync cell.

Marked the raw fns as `unsafe` and added a test case for regression. Test is a bit racy without the fix (may panic or segfault depending on sequencing), but confirmed we get
```
process didn't exit successfully: `... race_multiple_threads_to_launch_jvm --exact --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```
without the fix and ok with it.

### DIFF
--- a/src/jvm.rs
+++ b/src/jvm.rs
@@ -211,19 +211,35 @@ impl JvmBuilder {
             crate::libjvm::libjvm_or_load_at(&path)?;
         }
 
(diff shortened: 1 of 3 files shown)


nikomatsakis (approved):
We'll have to do a more careful review later.

Issue #41, opened 2023-04-28
model `Class` type parameters


nikomatsakis:
We currently model `Class` as an erased class. This is because we cannot support return types with wildcards, but `Class` is used very often as `Class<?>` (and in particular the `arrayType()` method, which we invoke internally, returns `Class<?>`).

PR #55, opened 2023-05-11
Improvements to use Scala's companion objects

fpoli:
General improvements:
* I extended the grammar in order to accept the `throws java/lang/IndexOutOfBoundsException` that `javap` generates, where the class reference uses `/` instead of `.`.
* `javap` outputs `public static {};` for classes with a static initialization block, also called `<clinit>`. I extended the grammar to parse and ignore that, since the initialization block is automatically executed by the JVM when necessary.
* I hit the issue https://github.com/duchess-rs/duchess/issues/41: in `java_package!` I needed to write `ArrayBuffer` but then the compiler was complaining that a type parameter was missing. So, I added a `java::lang::Object` default to *every* generated type parameter.

To support identifiers with `$` (https://github.com/duchess-rs/duchess/issues/12):
* Parsing classes with `$` was not supported in `java_package!`, because of whitespaces introduced when printing a token stream ([relevant comment](https://github.com/duchess-rs/duchess/issues/12#issuecomment-1535951998)). I extended the grammar to support identifiers in quotes. Example:
    ```
    class "NoopReporter$" implements viper.silver.reporter.Reporter {
        static viper.silver.reporter."NoopReporter$" "MODULE$";
    }
    ```
* When generating Rust identifiers, `Id::to_ident` was already replacing `$` with `__`, but that was not done consistently. So, I replaced all the `Ident::new(..)` that I could find with `Id::new(..).to_ident(..)`.

To use static fields:
* I added a `FieldPtr` to `raw.rs` and a `find_field` to `find.rs` (CC @chase-ok)
* I added the generation of `fn get_{field}()` Rust methods to obtain static Java fields.
* Side note: the `codegen.rs` file is getting very big and there are various repetitions (e.g. `static_method`/`object_method`/`static_field_getter`, and `jni_call_fn`/`jni_static_call_fn`/`jni_static_field_get_fn`/...). I didn't investigate how to refactor that.

Testing:
* I added a few tests to `test-crates/viper`. The most annoying part is that to call the inherited `to_seq` method on a Scala's `ArrayBuffer` I had to write ~300 characters of upcasts.

### DIFF
--- a/macro/src/codegen.rs
+++ b/macro/src/codegen.rs
@@ -1,8 +1,8 @@
 use crate::{
     argument::DuchessDeclaration,
     class_info::{
-        ClassInfo, ClassRef, Constructor, DotId, Generic, Id, Method, NonRepeatingType, RefType,
-        RootMap, ScalarType, SpannedPackageInfo, Type,
+        ClassInfo, ClassRef, Constructor, DotId, Field, Generic, Id, Method, NonRepeatingType,
+        RefType, RootMap, ScalarType, SpannedPackageInfo, Type,
     },
     reflect::Reflector,
     span_error::SpanError,
@@ -55,7 +55,7 @@ impl SpannedPackageInfo {
         reflector: &mut Reflector,
     ) -> Result<TokenStream, SpanError> {
         let package_id = DotId::new(parents, &self.name);
-        let name = Ident::new(&self.name, self.span);
+        let name = self.name.to_ident(self.span);
 
         let subpackage_tokens: TokenStream = self
             .subpackages
@@ -96,6 +96,7 @@ impl ClassInfo {
         let ext_trait_name = self.ext_trait_name();
         let cached_class = self.cached_class();
         let this_ty = self.this_type();
+        let java_class_generics_with_defaults = self.class_generic_names_with_defaults();
         let java_class_generics = self.class_generic_names();
 
         // Convert constructors
@@ -105,7 +106,7 @@ impl ClassInfo {
             .map(|c| self.constructor(c))
             .collect::<Result<_, _>>()?;
 
-        // Convert class methods (not static methods, those are different)
+        // Convert static methods (not instance methods, those are different)
         let static_methods: Vec<_> = self
             .methods
             .iter()
@@ -113,12 +114,20 @@ impl ClassInfo {
             .map(|m| self.static_method(m))
             .collect::<Result<_, _>>()?;
 
-        // Convert class methods (not static methods, those are different)
+        // Convert instance methods (not static methods, those are different)
         let object_methods: Vec<_> = self
             .methods
             .iter()
             .filter(|m| !m.flags.is_static)
             .map(|m| self.object_method(m))
+            .collect::<Result<_, _>>()?;
+
+        // Generate static field getters
+        let static_field_getters: Vec<_> = self
+            .fields
+            .iter()
+            .filter(|f| f.flags.is_static)
+            .map(|f| self.static_field_getter(f))
             .collect::<Result<_, _>>()?;
 
         let method_structs: Vec<_> = object_methods.iter().map(|m| &m.method_struct).collect();
@@ -134,7 +143,7 @@ impl ClassInfo {
             self.span =>
 
             #[allow(non_camel_case_types)]
-            pub struct #struct_name<#(#java_class_generics,)*> {
+            pub struct #struct_name<#(#java_class_generics_with_defaults,)*> {
                 _dummy: std::marker::PhantomData<(#(#java_class_generics,)*)>
             }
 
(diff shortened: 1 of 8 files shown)


test-crates/viper/tests/test.rs
```diff
@@ -1,17 +1,69 @@
...
+use viper::viper::silver::reporter::NoopReporter__;
 
-static INIT: Once = Once::new();
+/// Builds an empty Scala Seq of the specified type
+fn empty_scala_seq<T: duchess::JavaObject>() -> impl IntoJava<Seq<T>> {
+    ArrayBuffer::new()
+        .upcast::<StrictOptimizedSeqOps<T, ArrayBuffer, ArrayBuffer<T>>>()
+        .upcast::<SeqOps<T, ArrayBuffer, ArrayBuffer<T>>>()
+        .upcast::<IterableOps<T, ArrayBuffer, ArrayBuffer<T>>>()
+        .upcast::<IterableOnceOps<T, ArrayBuffer, ArrayBuffer<T>>>()
```
nikomatsakis, line 21: This is because we are not generating (yet) the transitive upcasts? I've been meaning to drop some mentoring notes in the issue on that. I think if we did that properly, you wouldn't need to generate *any* of these.

There is a bit of complication though in that generating transitive upcasts isn't always possible right now.

macro/src/codegen.rs
```diff
@@ -744,6 +857,13 @@ impl ClassInfo {
             .collect()
     }
 
+    fn class_generic_names_with_defaults(&self) -> Vec<TokenStream> {
+        self.class_generic_names()
+            .into_iter()
+            .map(|g| quote_spanned!(self.span => #g = java::lang::Object))
```
nikomatsakis, line 863: Interesting. The downside of this is that it gives us inaccurate return types, right?

Previously I was considering adding a special struct like `UnknownClass` or `Wildcard` and then translating `?` in return position to that struct. It's not *100%* correct -- but it's probably close enough for "government work". It would allow you to call Java methods that perhaps you ought not to be able to call (e.g., a method like `<T> (Class<T>, Class<T>)`), but is that a realistic danger?

Anyway, that seems preferable to using `java::lang::Object` here...? Maybe not always.

nikomatsakis (commented):
This all looks great except I don't quite understand the `Object` default yet.

test-crates/viper/src/lib.rs
```diff
@@ -1,21 +1,73 @@
...
+    interface IterableOps<A, CC, C> extends scala.collection.IterableOnceOps<A, CC, C> {}
+    interface SeqOps<A, CC, C> extends scala.collection.IterableOps<A, CC, C> {}
+    interface StrictOptimizedSeqOps<A, CC, C> extends scala.collection.SeqOps<A, CC, C> {}
+
     package scala.collection.immutable;
     interface Seq<A> {}
 
+    package scala.collection.mutable;
+    class ArrayBuffer<A> implements scala.collection.StrictOptimizedSeqOps<
+        A, scala.collection.mutable.ArrayBuffer, scala.collection.mutable.ArrayBuffer<A>
```
nikomatsakis, line 23: What is going on here? `ArrayBuffer` implements itself?

nikomatsakis: To start, I guess `ArrayBuffer` is an interface? We should really fix that.

test-crates/viper/tests/test.rs
nikomatsakis: Filed https://github.com/duchess-rs/duchess/issues/56

test-crates/viper/src/lib.rs
nikomatsakis: ...oh.

nikomatsakis: Oh, and it includes HKT. Fun.

fpoli: If I replace the first `ArrayBuffer` on line 23 with the semantically equivalent `ArrayBuffer<java.lang.Object>`, then duchess fails with "declared interface `...` not found in the reflected interfaces (`...`)" because the check is syntactic. That's why I looked into how to generate a default argument.

nikomatsakis:
OK, @fpoli, I think we should land this, but we should file an issue and discuss how to better handle erasure.

macro/src/codegen.rs
nikomatsakis: Filed #57

test-crates/viper/src/lib.rs
nikomatsakis: Hmm. It's not actually 100% semantically equivalent, although it doesn't matter here. That said, it seems *plausible* for us to extend the check?

nikomatsakis: That could be a lot of work, I guess. Can you file an issue with some notes on this *particular* problem (it's a bit different than #57, I realize now) and the fact that you added defaults?

nikomatsakis (approved):

PR #60, opened 2023-05-12
Implement basic derive support

### LINKED ISSUE #58: derive ToRust, JvmOp, and ToJava for structs
I'm currently working on a derive for structs to replace some of the boilerplate code in auth. Basic idea

```rust
#[derive(duchess::Java)]
#[java(some.java.Class)]
struct Foo {
    /* fields must correspond to arguments of constructor as well as accessor methods, unless there is a special `this` field */
}
```

we can work on improvements and extensions iteratively (as well as supporting enums).

nikomatsakis:
This adds two derives that work only on structs. Here is one example:

```rust
#[derive(duchess::ToRust, duchess::ToJava)]
#[java(auth.Authenticated)]
pub struct Authenticated {
    pub account_id: String,
    pub user: String,
    this: Global<auth::Authenticated>,
}
```

It's pretty basic right now. You specify with `#[java]` the class / constructor method you want. The field names have to be in the same order as the constructor arguments, and their types have to be compatible with `to_java`. The `this` field, if present, stores original Java object. It also requires that users have imported java names. 

Fixes #58


examples/auth.rs
```diff
@@ -1,67 +1,65 @@
...
 }
 
+#[derive(duchess::ToRust, duchess::ToJava)]
+#[java(auth.Authenticated)]
 pub struct Authenticated {
     pub account_id: String,
     pub user: String,
-    this: Global<java_auth::Authenticated>,
+    this: Global<auth::Authenticated>,
 }
```
chase-ok, line 63: awesome sauce

macro/src/derive.rs
```diff
@@ -0,0 +1,361 @@
...
+use crate::{argument::MethodSelector, parse::Parser, reflect::Reflector, signature::Signature};
+
+pub fn derive_to_rust(s: synstructure::Structure) -> proc_macro2::TokenStream {
+    let mut driver = Driver {
+        input: &s,
+        reflector: &mut Reflector::default(),
+    };
+    match driver.try_derive_to_rust() {
+        Ok(t) => {
+            if let Ok(debug_env) = std::env::var("DUCHESS_DEBUG") {
```
chase-ok, line 18: [aside] is there a way in the macro system to have it dump the generated code with a standard debug/verbose option? Seems like a we have a lot of copies of this in places

macro/src/derive.rs
```diff
@@ -0,0 +1,361 @@
...
+        let construct = variant.construct(|_field, index| {
+            assert!(counter == index);
+            counter += 1;
+            initializers.pop_front().unwrap()
+        });
+
+        Ok(construct)
+    }
+
+    /// Generates the code to create this variant as part of a `ToRust` impl.
```
chase-ok, line 227: [minor] `ToJava` I assume?

macro/src/derive.rs
```diff
@@ -0,0 +1,361 @@
...
+        variant: &VariantInfo,
+    ) -> Result<proc_macro2::TokenStream, syn::Error> {
+        let variant_span = variant.ast().ident.span();
+
+        // If there is a field named `this`, just return that.
+        if let Some(binding) = variant
+            .bindings()
+            .iter()
+            .find(|b| b.ast().ident.as_ref().map(|i| i == "this").unwrap_or(false))
+        {
```
chase-ok, line 240: What does it mean to modify a field in Rust but then use `this` in Java land? It could lead to some unexpected behavior like
```rust
let mut value = compute_something_in_java();
assert_eq!(value.int_field, 2);
value.int_field = 4;
compute_something_else_in_java(&value); // is int_field 2 or 4 from Java's perspective?
```

I'd suggest if we have `this` field, we prevent mutable access to the fields through some getter fns.

chase-ok (approved):
LGTM aside from the mutability of fields when we also have a `this` field. Perhaps we can resolve that in a separate PR, though!

chase-ok (approved):
LGTM aside from the mutability of fields when we also have a `this` field. Perhaps we can resolve that in a separate PR, though!

macro/src/derive.rs
nikomatsakis:
```suggestion
    /// Generates the code to create this variant as part of a `ToJava` impl.
```

macro/src/derive.rs
nikomatsakis: Hmm. That's true. We could error if there are public fields and a `this` field, but of course it won't prevent people from doing the wrong thing. The other option (more drastic) would be to use a wrapper type that prevents direct writes and only gives `&` access. That seems a bit extreme to me. 

(In retrospect, I wish that Rust had you declare mutable fields as mutable... maybe worth proposing as an RFC sometime.)

PR #69, opened 2023-05-15
Add ui test suite

### LINKED ISSUE #5: test suite
We need to start setting up a proper test suite.

oli-obk:
fixes #5

I was unable to improve the error span further, as the duchess parser stringifies the class body before parsing with larlpop. We'd need to teach larlpop to handle token streams first.

Alternatively we can use the offset in the stringified version to walk the token stream to find the same offset and use that span.


nikomatsakis:
> We'd need to teach larlpop to handle token streams first.

Yes, I was thinking about doing that once LALRPOP gets going.

PR #71, opened 2023-05-16
instantiate transitive upcasts

### LINKED ISSUE #56: transitive upcasts
So both @fpoli (in #55) and @chase-ok (in the auth example) encountered the fact that we don't generate transitive upcasts. I've given some thought to how to handle this. There are a few complications. Here is a proposal of what to do "for now".

### Example

Here is a simple example that shows some of the complications

```rust
java_package! {
    package foo;
    class Foo extends bar.Bar {}

    package bar;
    class Bar extends bar.Baz {}
    class Baz extends java.util.Time {}
}
```

### Basic algorithm: enumerate transitive upcasts

The basic idea is that we will enumerate all the transitive upcasts that involve classes which are going to be translated within one `java_package` macro invocation. So we would generate, for `foo.Foo`, upcasts 

* `Foo: Upcast<Bar>`
* `Foo: Upcast<Baz>`
* `Foo: Upcast<java::util::Time>`
* `Foo: Upcast<java::lang::Object>` -- hard-coded, we always do this

but there are some subtle points to this algorithm.

### Subtle point 1: Only use the upcasts declared by user

It's possible that some of the types in the `java_package` have more upcasts than what was declared by the user. We should only use those upcasts they declared, because the omitted ones were probably omitted for a reason. For example, perhaps `Foo` also implements some Java interfaces related to reflection that we are not modeling. We already check that every name the user actually writes is included in `java_package`, so actually we don't need to use the reflection here at all, just consult the `ClassInfo` we parsed from what the user wrote.

### Subtle point 2: Enumeration stops at external types

Note that our list of upcasts stopped at `java::util::Time`. This is because it belongs to a package that is outside the list of packages we are translating, so we don't have access to its data. Because of the previous point, we can't rely on Java reflection as a source of truth, as it may include superclasses/interfaces that are not meant to be translated. Of course we can always implement `java::lang::Object`.

As a workaround, we can allow the user to write the transitive cases here themselves, for example:

```rust
java_package! {
    /* ... as above */
    class Baz extends 
        java.util.Time, 
        java.util.Date  // transitive case
    {}
}
```

This requires us to modify the correctness checking to permit superclasses/interfaces as long as they are part of the transitive closure (in contrast, we can compute THIS transitive closure using reflect, which is interesting and a bit wacky; but if it includes a type that is not translated, the generated Rust code will fail to build anyway).

### Subtle point 3: Substitution

I've been ignoring generics here, but we have to implement substitution. Should be fairly straightforward.

nikomatsakis:
The current code stops at things outside the package, with certain very specific exceptions from the JDK (which we may need to extend).

Fixes #56

### DIFF
--- a/macro/src/upcasts.rs
+++ b/macro/src/upcasts.rs
@@ -5,7 +5,7 @@ use crate::{
     substitution::{Substitute, Substitution},
 };
 
-/// A map storing the transitive upcasts for each class that we are generated (and potentially additional classes).
+/// A map storing the transitive upcasts for each class that we are generating (and potentially additional classes).
 ///
 /// There is one caveat: we only compute the transitive superclasses based on the classes that are input to
 /// the proc macro. The problem is that we can only inspect the tokens presented to us. While we could reflect
--- a/macro/src/upcasts.rs
+++ b/macro/src/upcasts.rs
@@ -83,6 +83,13 @@ impl Upcasts {
                 });
         };
 
+        // N.B. we don't insert the reflexive `C extends C` relations here, which means we don't
+        // necessarily have 100% parity between these hardcoded types and types created by
+        // `insert_direct_upcasts`. This is a micro-optimization that I can't resist.
+        // The idea is that reflexive impls aren't necessary because they only matter when
+        // we are generating `Upcast` impls, and we only generate `Upcast` impls for those
+        // classes that appear in our package declarations.
+
         insert(DotId::runtime_exception(), DotId::exception());
         insert(DotId::exception(), DotId::throwable());
         insert(DotId::throwable(), DotId::object());
--- a/macro/src/upcasts.rs
+++ b/macro/src/upcasts.rs
@@ -1,0 +1,137 @@
+use std::collections::{BTreeMap, BTreeSet};
+
+use crate::{
+    class_info::{ClassInfo, ClassRef, DotId, Id},
+    substitution::{Substitute, Substitution},
+};
+
+/// A map storing the transitive upcasts for each class that we are generated (and potentially additional classes).
+///
+/// There is one caveat: we only compute the transitive superclasses based on the classes that are input to
+/// the proc macro. The problem is that we can only inspect the tokens presented to us. While we could reflect
+/// on the Java classes directly, we don't know what subset of the supertypes the user has chosen to reflect into
+/// Rust. Therefore, we stop our transitive upcasts at the "water's edge" -- i.e., at the point where we
+/// encounter classes that are outside our package.
+#[derive(Default, Debug)]
+pub struct Upcasts {
+    map: BTreeMap<DotId, ClassUpcasts>,
+}
+
+#[derive(Debug)]
+pub struct ClassUpcasts {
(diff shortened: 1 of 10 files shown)


macro/src/upcasts.rs
```diff
@@ -0,0 +1,137 @@
...
+                .extends
+                .insert(ClassRef {
+                    name: d,
+                    generics: vec![],
+                });
+        };
+
+        insert(DotId::runtime_exception(), DotId::exception());
+        insert(DotId::exception(), DotId::throwable());
+        insert(DotId::throwable(), DotId::object());
```
fpoli, line 88: `insert_direct_upcasts` generates reflexive upcast relations, but the current `insert_hardcoded_upcasts` does not. I'm not sure if this is intended.

For example, if `compute_transitive_upcasts` starts with `{ E -> [T], RE -> [E], T -> [O] }` where `RE` stands for `RuntimeException` etc the current execution is:
* first loop iteration:
    * the inner loop iteration `n = E` ends with `{ E -> [T, O], RE -> [E], T -> [O] }`
    * the inner loop iteration`n = RE` ends with `{ E -> [T, O], RE -> [E, T, O], T -> [O] }`
    * the inner loop iteration`n = T` ends with `{ E -> [T, O], RE -> [E, T, O], T -> [O] }`
* second loop iteration:
    * nothing changes; the loop terminates.
* result: `E <: E`, `RE <: RE`, `T <: T` and `O <: O` are not in the map.

To make it consistent:
```suggestion
        insert(DotId::runtime_exception(), DotId::runtime_exception());
        insert(DotId::runtime_exception(), DotId::exception());
        insert(DotId::exception(), DotId::exception());
        insert(DotId::exception(), DotId::throwable());
        insert(DotId::throwable(), DotId::throwable());
        insert(DotId::throwable(), DotId::object());
        insert(DotId::object(), DotId::object());
```
Alternatively, the closure should also insert the reflexive relations for both arguments.

fpoli (approved):
I think a unit test for `Upcasts::from_iter` would be useful.

macro/src/upcasts.rs
nikomatsakis: It's intentional, but not harmful to include them. The reason is that the reflexive relations only matter when you are generating the `Exception` etc types, and in that case, you will get the reflexive relation from the `insert_direct_upcasts` call

nikomatsakis: I couldn't bring myself to do the clones needed to include them.

nikomatsakis:
> I think a unit test for Upcasts::from_iter would be useful.

Hmm. I generally agree and felt bad about the lack of unit tests in this PR. OTOH, they're a bit of a pain to author and the code felt "obviously right" to me (famous last words), and I only have very limited hacking time. Let me think if there's some easy way to set them up. 

Side not but I'd also prefer to avoid "unit tests" per se and instead write integration tests using the UI testing framework.

nikomatsakis:
I pushed one test, the viper commit. The problem is that the current JVM types we model are not sufficient to really test the code as we have such a shallow type hiearchy.

PR #73, opened 2023-05-18
Inherent methods

### LINKED ISSUE #68: resolve ambiguity between interfaces, supertypes, and subtypes
If you have a class `C` that implements an interface`I`, especially when using `*`, the same method appears in both the class trait `CExt` the interface trait `IExt` and creates an ambiguity error. This shows up in the `log` example -- if you import all the `Ext` traits, you'll get ambiguity errors.

I'm not sure the best way to fix this. My current plan is to exclude methods from `*` if the same method appears in an interface or superclass (or perhaps just one that is *also* being generated?). That is not the best solution for a few reasons. Another option would be to trick the compiler by leveraging autoderef, but I haven't quite figured out how to do that yet, particularly since we support methods being implemented on other `JvmOp`. I will think about that.

nikomatsakis:
This PR implements (or rather, will implement once it is complete) a new approach to handling method resolution order. For now, it consists only of a mdbook chapter describing my plan, as that is all I had time for. 

Fixes #68


nikomatsakis:
OK @duchess-rs/devs this is ready to go!

There are two commits. The first one documents the design in some depth. The second one implements it. I had more but the history was so messy that I didn't feel it was helping anybody.

The main thing missing, I think, are tests for the edge cases. I don't have time for that right now but I'm going to file an issue about it with some of the cases I can think of.

PR #78, opened 2023-05-28
Derive `ToRust` and `ToJava` for enums

### LINKED ISSUE #61: extend derive with support for enums
#60 includes support for deriving `ToJava` and `ToRust` for structs, but not enums. 

For enums, the plan I had was:

* `#[java(class.name)]` on the enum itself
* optional `#[java(subclass.name)]` on each variant, which is then treated exactly like the code for a struct (and the code should mostly be ready for that)
* we will check that each subclass for the variants extends the enum class, and that there is exactly one variant without a `#[java]` annotation (which serves as a "fall through").

When we implement `ToRust`, we will check whether the input is a subtype of the class associated with each variant and use that constructor if so, falling back to the one variant that has no `#[java]` annotation. This will use the same `variant_to_rust` code as for structs.

chase-ok:
Extends our existing `ToRust` and `ToJava` derive macros to handle enum cases. For the most part, this is straightforward. The few tricky cases include:
1. Making sure that when we're deriving `ToRust` that we perform `instanceof` checks in order from most specific to least (i.e. a topological sort order). This prevents us from missing subclass variants.
2. Enforce that all variants descend from a common root class, which must be specified as a plain class name.
3. Upcast the results of more specific variants for `ToJava` enums. 

I've also removed the use of `reflect_method()` for `ToRust` cases because we don't actually need an unambiguous Java constructor if all we're deriving is the to-Rust direction.

Closes #61


nikomatsakis (approved):
This is awesome! We need more tests, but I'm not going to block the PR on it.

PR #93, opened 2023-06-22
Native functions

nikomatsakis:
Initial implementation of native functions.

First commit documents the plan.

~~This PR needs a rebase, apparently, but also it doesn't work!~~

Works now!

### DIFF
--- a/src/to_java.rs
+++ b/src/to_java.rs
@@ -1,9 +1,12 @@
 use std::{collections::HashMap, marker::PhantomData};
 
-use crate::{cast::Upcast, java, Error, Global, Jvm, JvmOp, Local};
+use crate::{
+    cast::Upcast, from_ref::FromRef, java, jvm::JavaView, Error, Global, Jvm, JvmOp, Local,
+};
 
 pub trait ToJava: Sized {
     type JvmOp<'a, J>: for<'jvm> JvmOp<Output<'jvm> = Option<Local<'jvm, J>>>
+        + std::ops::Deref<Target = <J as JavaView>::OfOp<Self::JvmOp<'a, J>>>
     where
         Self: 'a,
         Self: ToJavaImpl<J>,
@@ -59,6 +62,18 @@ where
 
     fn execute_with<'jvm>(self, jvm: &mut Jvm<'jvm>) -> crate::Result<'jvm, Self::Output<'jvm>> {
         R::to_java_impl(self.rust, jvm)
+    }
+}
+
+impl<R, J> std::ops::Deref for ToJavaOp<'_, R, J>
+where
+    R: ToJavaImpl<J>,
+    J: Upcast<java::lang::Object> + Upcast<J>,
+{
+    type Target = <J as JavaView>::OfOp<Self>;
+
+    fn deref(&self) -> &Self::Target {
+        <Self::Target as FromRef<_>>::from_ref(self)
     }
 }
 
--- a/tests/ui/to_java_java_refs.rs
+++ b/tests/ui/to_java_java_refs.rs
@@ -10,33 +10,41 @@ use duchess::{Global, Local};
 // Test that `to_java` can accomodate a Rust vector of (local) Java objects
 // and produce a Java list of Java objects.
 fn produce_from_local_rust_vec(r: &Vec<Local<'_, JavaString>>) {
-    let _data: Option<Global<JavaList<JavaString>>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaList<JavaString>>> = r
+        .to_java::<JavaList<JavaString>>()
+        .global()
+        .execute()
+        .unwrap();
 }
 
 // Test that `to_java` can accomodate a Rust vector of (global) Java objects
 // and produce a Java list of Java objects.
 fn produce_from_global_rust_vec(r: &Vec<Global<JavaString>>) {
-    let _data: Option<Global<JavaList<JavaString>>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaList<JavaString>>> = r
+        .to_java::<JavaList<JavaString>>()
+        .global()
+        .execute()
+        .unwrap();
 }
 
 // Test that `to_java` can accomodate a global Java object.
 fn produce_from_global_object(r: Global<JavaString>) {
-    let _data: Option<Global<JavaString>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaString>> = r.to_java::<JavaString>().global().execute().unwrap();
 }
 
 // Test that `to_java` can accomodate a local Java object.
 fn produce_from_local_object(r: Local<'_, JavaString>) {
-    let _data: Option<Global<JavaString>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaString>> = r.to_java::<JavaString>().global().execute().unwrap();
 }
 
 // Test that `to_java` can accomodate an optional local Java object.
 fn produce_from_optlocal_object(r: Option<Local<'_, JavaString>>) {
-    let _data: Option<Global<JavaString>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaString>> = r.to_java::<JavaString>().global().execute().unwrap();
 }
 
 // Test that `to_java` can accomodate a ref to an optional local Java object.
 fn produce_from_optlocal_object_ref(r: &Option<Local<'_, JavaString>>) {
-    let _data: Option<Global<JavaString>> = r.to_java().global().execute().unwrap();
+    let _data: Option<Global<JavaString>> = r.to_java::<JavaString>().global().execute().unwrap();
 }
 
 fn main() {}
--- a/tests/ui/to_java_on_java_string.rs
+++ b/tests/ui/to_java_on_java_string.rs
@@ -1,0 +1,20 @@
+//@ run
+
+#![allow(dead_code)]
+
+use duchess::java;
+use duchess::prelude::*;
+use duchess::Global;
+
+fn main() -> duchess::GlobalResult<()> {
+    let data: Global<java::lang::String> = format!("Hello, Duchess!").global().execute()?;
+
+    let hash_code: i32 = data
+        .to_java::<java::lang::String>() // Returns a `JvmOp` producing a `java::lang::String`
+        .hash_code() // Returns a `JvmOp` invoking `hashCode` on this string
+        .execute()?; // Execute the jvmop
+
+    // NB: [hashCode for string is documented](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#hashCode--)
+    assert_eq!(hash_code, -928531272);
+    Ok(())
+}
--- a/tests/ui/to_java_on_rust_string.rs
+++ b/tests/ui/to_java_on_rust_string.rs
@@ -1,0 +1,18 @@
+//@ run
+
+#![allow(dead_code)]
+
+use duchess::java;
+use duchess::prelude::*;
+
+fn main() -> duchess::GlobalResult<()> {
+    let data = format!("Hello, Duchess!");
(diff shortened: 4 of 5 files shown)


src/jvm.rs
```diff
@@ -125,6 +133,104 @@ fn get_or_default_init_jvm() -> crate::GlobalResult<JvmPtr> {
...
+    let old_state = thread::attach_from_jni_callback(env);
+
+    let result = match std::panic::catch_unwind(AssertUnwindSafe(|| op())) {
+        Ok(result) => {
+            let mut jvm = Jvm(env);
+            eprintln!("got result = {result:?}");
+            let obj = result.to_java().execute_with(&mut jvm);
+            eprintln!("obj = {obj:?}");
+            match obj {
+                Ok(Some(p)) => p.as_raw().as_ptr(),
```
fpoli, line 162: Here, isn't `obj: Local` deallocated on `Drop` when this Rust function terminates, while it should instead remain allocated to be used from Java? I guess that if the returned object points to a deallocated instance then the JVM might conservatively interpret it as being `null`.

nikomatsakis: *Ah*, yeah, that's prob it, interesting!

nikomatsakis: Good catch @fpoli, I'll tweak that

nikomatsakis:
Hey @fpoli -- have time to give this a review now that it's no longer in draft status?

src/jvm.rs
```diff
@@ -125,6 +133,100 @@ fn get_or_default_init_jvm() -> crate::GlobalResult<JvmPtr> {
...
+///
+/// Must be invoked as the entire body of a JNI native function, with
+/// `env` being the `EnvPtr` argument provided.
+pub unsafe fn native_function_returning_object<J, R>(
+    env: EnvPtr<'_>,
+    op: impl FnOnce() -> R,
+) -> jni_sys::jobject
+where
+    J: Upcast<crate::java::lang::Object> + Upcast<J>,
+    R: ToJavaImpl<J> + std::fmt::Debug,
```
fpoli, line 150: Do you still need the `Debug`?
```suggestion
    R: ToJavaImpl<J>,
```

fpoli:
Is it possible to return Java objects from a native function? I tried changing the `native_greetings.rs` test to use the following native function, but it's rejected.
```rust
#[duchess::java_function(native_greeting.Native::baseGreeting)]
fn base_greeting(
    _this: &native_greeting::Native,
    name: &java::lang::String,
) -> duchess::GlobalResult<java::lang::String> {
    todo!()
}
```
The error is:
```
   --> tests/ui/native_greeting.rs:15:50
    |
15  | #[duchess::java_function(native_greeting.Native::baseGreeting)]
    |                                                  ^^^^^^^^^^^^ the trait `ToJavaImpl<duchess::java::lang::String>` is not implemented for `duchess::java::lang::String`
```

fpoli:
```rust
#[duchess::java_function(native_greeting.Native::baseGreeting)]
fn base_greeting(
    env: &mut JvmEnv<'_>, // Optional?
    _this: &native_greeting::Native,
    name: &java::lang::String,
) -> duchess::GlobalResult<String> {
    let _ = env;
    let name: String = name.to_rust().execute()?;
    Ok(format!("Hello, {name}"))
}
```
fails to compile with
```
error: extra argument(s) on Rust function, only 2 argument(s) are expected
  --> tests/ui/native_greeting.rs:19:5
   |
19 |     name: &java::lang::String,
   |     ^^^^
```
The documentation explains that `env` is optional, so I didn't expect this error.

book/src/java_object.md
```diff
@@ -0,0 +1 @@
+# The `JavaObject` trait
```
fpoli, line 1: Why this empty page? Is it linked from elsewhere in the book?

book/src/linking_native_functions.md
```diff
@@ -0,0 +1,96 @@
...
+**How it works.** The call `foo::java_fn()` returns a `duchess::JavaFunction` struct. The `java_fn` method is defined in the duchess `JavaFn` trait; that trait is implemented on a struct type `foo` that is created by the `#[java_function]` decorator. This trait is in the duchess prelude, which is why you need to `use duchess::prelude::*`.
+
+### Java function suites
+
+Invoking the link method for every java functon you wish to implement is tedious and error-prone. If you have java functions spread across crates and modules, it also presents a maintenance hazard, since each time you add a new `#[java_function]` you would also have to remember to add it to the Jvm builder invocation, which is likely located in some other part of the code.
+
+To avoid this, you can create **suites** of java functions. The idea is that the `link` method accepts both individual `JavaFunction` structs but also `Vec<JavaFunction>` suites. You can then write a function in your module that returns a `Vec<JavaFunction>` with all the java functions defined locally:
+
+```rust
+use duchess::prelude::*;
```
fpoli, line 31: I think `JavaFunctions` is not in the prelude, because I tried defining a `java_functions()` in the `native_greetings.rs` test and I get the following:
```
error[E0412]: cannot find type `JavaFunction` in this scope
  --> tests/ui/native_greeting.rs:24:28
   |
24 | fn java_functions() -> Vec<JavaFunction> {
   |                            ^^^^^^^^^^^^ not found in this scope
   |
help: consider importing this struct
   |
3  | use duchess::plumbing::JavaFunction;
   |
```

book/src/linking_native_functions.md
```diff
@@ -0,0 +1,96 @@
...
+        foo::java_fn(),
+        bar::java_fn(),
+    ]
+}
+```
+
+You can also compose suites from other crates or modules:
+
+```rust
+fn java_functions() -> Vec<duchess::JavaFunction> {
```
fpoli, line 50: For consistency with the other snippets
```suggestion
fn java_functions() -> Vec<JavaFunction> {
```

book/src/linking_native_functions.md
```diff
@@ -0,0 +1,96 @@
...
+
+fn main() -> duchess::GlobalResult<()> {
+    Jvm::builder()
+        .link(foo::java_fn()) // 👈 Note the `::java_fn()`.
+        .try_launch()?;
+}
+```
+
+**How it works.** The call `foo::java_fn()` returns a `duchess::JavaFunction` struct. The `java_fn` method is defined in the duchess `JavaFn` trait; that trait is implemented on a struct type `foo` that is created by the `#[java_function]` decorator. This trait is in the duchess prelude, which is why you need to `use duchess::prelude::*`.
+
```
fpoli, line 23: Linking the same function multiple times seems to work fine. If it's intended, should this be mentioned in the documentation?

book/src/to_java.md
```diff
@@ -0,0 +1,49 @@
...
+### `String`
+
+The Rust `String` type converts to the Java string type.
+One could compute the Java `hashCode` for a string as follows:
+
+```rust
+let data = format!("Hello, Duchess!");
+let hashCode: i32 =
+    data.to_java()   // Returns a `JvmOp` producing a `java::lang::String`
+        .hashCode()  // Returns a `JvmOp` invoking `hashCode` on this string
```
fpoli, line 26: `public int hashCode();` is not declared for `java.lang.String` in `src/java.rs`.

Could you double check this snippet? If I try to compile it (replacing `hashCode` with `length`) I get the following error, but I'm not sure if it's because I miss some import or if it's because the chaining is still unimplemented.
```
error[E0599]: no method named `length` found for struct `ToJavaOp` in the current scope
  --> tests/ui/native_greeting.rs:21:40
   |
21 |     let nameJava: i32 = name.to_java().length().execute();
   |                                        ^^^^^^ method not found in `ToJavaOp<'_, String, String>`
```

java/native_greeting/Native.java
```diff
@@ -0,0 +1,9 @@
+package native_greeting;
+
+public class Native {
+    public String greet(String name) {
+        return baseGreeting(name) + ", from Java";
+    }
+
+    native String baseGreeting(String name);
+}
```
fpoli, line 9:
```suggestion
}

```

src/error.rs
```diff
@@ -89,12 +85,5 @@ impl<'jvm> Error<Local<'jvm, Throwable>> {
 pub fn check_exception<'jvm>(jvm: &mut Jvm<'jvm>) -> Result<'jvm, ()> {
     let env = jvm.env();
     // SAFETY: we don't hold on to the return env ptr
```
fpoli, line 87:
```suggestion
    // SAFETY: we don't hold on to the return env ptr
```

src/thread.rs
```diff
@@ -41,6 +42,23 @@ fn attached_or(
...
+pub unsafe fn detach_from_jni_callback(env: EnvPtr<'_>, s0: State) {
+    let env: EnvPtr<'static> = unsafe { std::mem::transmute(env) };
+    STATE.with(|state| {
+        let s1 = state.replace(s0);
+        assert!(
+            s1 == State::AttachedPermanently(env),
+            "invalid prior state `{s1:?}`"
+        );
+    });
+}
```
fpoli, line 60: What's the safety condition for these two functions?

src/jvm.rs
```diff
@@ -125,6 +133,100 @@ fn get_or_default_init_jvm() -> crate::GlobalResult<JvmPtr> {
...
+    let old_state = thread::attach_from_jni_callback(env);
+
+    let result = match std::panic::catch_unwind(AssertUnwindSafe(|| op())) {
+        Ok(result) => {
+            let mut jvm = Jvm(env);
+            let obj = result.to_java().execute_with(&mut jvm);
+            match obj {
+                Ok(Some(p)) => p.into_raw().as_ptr(),
+                Ok(None) => std::ptr::null_mut(),
+                Err(e) => todo!("convert from exception `{e:?}` into java exception"),
```
fpoli, line 162: The `todo!` (here an inside `rust_panic_to_java_exception`) will skip the `detach_from_jni_callback`, breaking the invariant of `STATE`. I wonder how bad that is. Maybe some safety condition is violated as well.

fpoli (approved):
The comment about panics in `native_function_returning_object`/`native_function_returning_scalar` is the only big one. All other comments are minor.

src/jvm.rs
nikomatsakis: Oh, I do not, good catch.

src/jvm.rs
fpoli: A type that calls `detach_from_jni_callback` on `Drop` might be useful, instad of manually checking all possible unwind paths.

book/src/java_object.md
nikomatsakis: I think this trait was going to have a bigger role than it would up having...? Probably an ok page to have though.

book/src/linking_native_functions.md
nikomatsakis: D'oh. Fixed. It's not in the *prelude* but it is in the crate outside of `plumbing` now.

Side note but I wonder if we can find a better way to deal with `plumbing`. It'd be cool if you could annotate something so the compiler wouldn't offer it in suggestions; or perhaps if the compiler would just avoid "doc-hidden" suggestions.

book/src/to_java.md
nikomatsakis: Yep. We really gotta get the snippets working to the point where they can be auto-compiled.

src/jvm.rs
nikomatsakis: Great suggestion! Done.

book/src/to_java.md
nikomatsakis: Oh, I see the problem, there's no Deref impl for these ...

nikomatsakis:
@fpoli 

Addressed feedback:

* [x] removed Debug requirement
* [x] test for returning a java object
* [x] removed doc of `env` as it is not implemented 
* [x] inline comments

nikomatsakis:
Thanks @fpoli for the detailed feedback! That really helped catch a lot of oversights.

PR #98, opened 2023-06-26
Allow running just tests whose path matches a specific string.

oli-obk:
Setting the `TESTNAME` env var while running tests will only run ui tests with that filename in their name.

### DIFF
--- a/tests/ui.rs
+++ b/tests/ui.rs
@@ -18,5 +18,24 @@ fn main() -> color_eyre::eyre::Result<()> {
     // Make sure we can depend on duchess itself in our tests
     config.dependencies_crate_manifest_path = Some("Cargo.toml".into());
 
-    run_tests(config)
+    let test_name = std::env::var_os("TESTNAME");
+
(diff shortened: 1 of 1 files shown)


tests/ui.rs
```diff
@@ -18,5 +18,24 @@ fn main() -> color_eyre::eyre::Result<()> {
     // Make sure we can depend on duchess itself in our tests
     config.dependencies_crate_manifest_path = Some("Cargo.toml".into());
 
-    run_tests(config)
+    let test_name = std::env::var_os("TESTNAME");
```
nikomatsakis, line 21: Useful, though using an env var feels vaguely annoying. I feel like this should be built into the framework.

nikomatsakis (approved):

tests/ui.rs
fpoli: It could also be `let test_name = env::args().nth(1);`, to enable `cargo test test-name`.

PR #100, opened 2023-07-05
Use `$JAVA_HOME/bin/javap` if `$JAVA_HOME` is set

chase-ok:
Some systems have multiple Java versions present and it's easy for the `javap` from one JDK to end up earlier on the path than the `javap` we're building against. The `JAVA_HOME` env var is a canonical way to identify and pass the JDK/JRE location to programs. We already use it for locating `libjvm` at runtime! If no `JAVA_HOME` was set, we fall back to relying on `PATH`, like we do for `libjvm`.

Updated the setup docs as well.

### DIFF
--- a/book/src/setup.md
+++ b/book/src/setup.md
@@ -1,10 +1,8 @@
 # Setup instructions
 
-## JDK
+## JDK and JAVA_HOME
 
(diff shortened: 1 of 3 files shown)


dchenbecker:
LGTM!

nikomatsakis (approved):

Issue #50, opened 2023-05-06
nested classes in java-package declaration


nikomatsakis:
We need to support things like

```java
package p;
class Foo { class Bar { } }
```

and then in Rust

```rust
duchess::java_package! {
    package p;
    class Foo.Bar { * }
}
```

nikomatsakis:
Not entirely clear how we should be mapping those... `Foo_Bar`? `foo::Bar`?

fpoli:
To support nested classes duchess should also need to parse types like `scala.Option<A>.WithFilter`, where the element in between the `.` is a *class name with type parameters* and not a package. The grammar used by the parser currently does not support type parameters before a dot.

The example comes from the return type of Scala's [`Option::withFilter`](https://www.scala-lang.org/api/2.12.5/scala/Option.html#withFilter(p:A=%3EBoolean):Option.this.WithFilter) method:
```
  public final scala.Option<A>.WithFilter withFilter(scala.Function1<A, java.lang.Object>);
    descriptor: (Lscala/Function1;)Lscala/Option$WithFilter;
```

PR #102, opened 2023-07-13
Parse (and ignore) generics in the middle of ClassRef

fpoli:
Duchess doesn't accept the following class declaration:
```rust
java_package!{
    package scala;
    class "None$" {}
}
```
The reason is that the reflection checks of Duchess try to parse the  `javap` output of `scala.None$`, which includes a method declaration that returns a type `scala.Option<A>.WithFilter`. The parser doesn't like the `<A>` in the middle of the type. This happens because `WithFilter` is an inner class of `scala.Option<A>`.

This PR makes the parser accept **and ignore** the type arguments in the middle of the path of a type. Since Duchess doesn't currently support nested classes, ignoring those arguments seems acceptable. As a result, the opaque class declaration above works.

I see two future possible improvements:
1. When the user declares an opaque class (i.e. with an empty body), the reflection checks of Duchess should just parse the *class signature* in the output of `javap` (i.e. everything that is before `{`), not also its body. This way, Duchess would be more robust.
2. Parsing `scala.Option<A>.WithFilter` should not throw away the information that `scala.Option` has type argument `A`. This is necessary to eventually support nested classes (https://github.com/duchess-rs/duchess/issues/50).


nikomatsakis (approved):

nikomatsakis:
Hmm, ok. Let's do this. Sorry for not noticing this for so long, @fpoli !

I agree with your improvements. Maybe open an issue for that first one in particular?

PR #109, opened 2023-07-28
Refer to `duchess` and `core` types explicitly

### LINKED ISSUE #108: Cannot declare `scala.Option`
Suppose a class with the name `Option` is declared, e.g.:

```rust
duchess::java_package! {
    package scala;
    class Option<A> {
        public A get();
    }
}
```

This fails with some difficult-to-read errors referring to the generated code, but this note in particular:

```
   = note: `std::option::Option<duchess::Local<'_, A>>` and `scala::Option<duchess::Local<'_, A>>` have similar names, but are actually distinct types
note: `std::option::Option<duchess::Local<'_, A>>` is defined in crate `core`
  --> /rustc/4c8bb79d9f565115637cc6da739f8389e79f3a29/library/core/src/option.rs:563:1
note: `scala::Option<duchess::Local<'_, A>>` is defined in the current crate
```

Makes me think the generated code should always refer to the declared type as `scala::Option` (or something relative to `self` perhaps) to avoid shadowing the stdlib `Option` or vice versa.

Aurel300:
Fixes #108.

All `std::*` paths changed to `::core::*`. All `duchess` types imported in codegen are now referred to through `duchess::*` paths. Trait method calls are spelled out to avoid importing the trait.

### DIFF
--- a/macro/src/codegen.rs
+++ b/macro/src/codegen.rs
@@ -142,20 +142,13 @@ impl ClassInfo {
 
             #[allow(non_camel_case_types)]
             pub struct #struct_name<#(#java_class_generics_with_defaults,)*> {
-                _dummy: std::marker::PhantomData<(#(#java_class_generics,)*)>
+                _dummy: ::core::marker::PhantomData<(#(#java_class_generics,)*)>
             }
 
             // Hide other generated items
             #[allow(unused_imports)]
             #[allow(nonstandard_style)]
             const _: () = {
-                use duchess::{
-                    *,
-                    plumbing::once_cell::sync::OnceCell,
-                    plumbing::*,
-                    prelude::*,
-                };
-
                 #assoc_struct_declarations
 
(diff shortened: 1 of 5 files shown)


fpoli (changes requested):
Could you declare `scala.Option` in `test-crates/viper/src/lib.rs` to avoid future regressions?

fpoli (approved):
Thanks!

Aurel300:
Ah, of course, adding `scala.Option` to the test crate means we need #102.

fpoli:
> Ah, of course, adding `scala.Option` to the test crate means we need #102.

Oh, I didn't foresee that. If you want to remove the `scala.Option` we can merge immediately; the changes are small anyway. Otherwise, we'll wait for #102.

Aurel300:
Ok, I removed `scala.Option`—let's add that in #102. For now this tests at least that `Result` is not clashing.

fpoli (approved):

PR #110, opened 2023-08-05
Generate test class files via build.rs

### LINKED ISSUE #88: build java files in build.rs
We should really build the java files used by tests and examples in a `build.rs` step instead of checking in `.class` files into the repository.

dchenbecker:
Add a custom build script to detect changes and auto-build class files for Java tests and examples. We assume that javap is already in the path, so we think it's safe to also assume javac is available. Also update the cargo CLASSPATH to point to the new generated directory and clean up the previously checked-in class files.

Fixes #88


fpoli:
Is the new `build.rs` executed whenever a crate has duchess as a dependency?

dchenbecker:
I'm honestly not sure. Is there an easy way to test that within the existing codebase or should I build a dummy crate to see if the transitive dependency works?

build.rs
```diff
@@ -0,0 +1,80 @@
...
+};
+
+use walkdir::WalkDir;
+
+// These two control where this script looks for source and corresponding class files
+const SOURCE_PATH: &str = "java";
+const TARGET_PATH: &str = "target";
+
+// Controls where javac puts generated class files
+const JAVAC_TARGET_PATH: &str = "target/java";
```
fpoli, line 13: The `/` separator might be a problem on Windows. Could you use something like [`std::path::MAIN_SEPARATOR`](https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html)?

dchenbecker: Yup, I'll just switch to using `join` to avoid the issue entirely.

dchenbecker:
OK, building `test-crates` does cause the Java files to get built, so are you suggesting that I make a new crate under `test-crates`, move the java sources, associated tests under `ui`, and the `examples` directory into that new crate?

fpoli:
> OK, building `test-crates` does cause the Java files to get built, so are you suggesting that I make a new crate under `test-crates`, move the java sources, associated tests under `ui`, and the `examples` directory into that new crate?

Yes, exactly. A second alternative is to make the programs in `examples` become `ui` tests, compiling the Java code only as an initialization step of the `ui.rs` test suite. Among the two I have a weak preference for first solution, the crate in `test-crates`, because it's probably closer to what real crates that depend on duchess need to do.

In both cases, it might be necessary to add `ignore` to some examples of the book to supporess compilation errors, because the CI currently [tests](https://github.com/duchess-rs/duchess/blob/43c6bc5afe77f3ea4007a7ff8e86679d02b31055/.github/workflows/rust.yml#L32) them and some seem to refer to classes defined in the Java files. Moreover, by removing the `examples` folder I expect that the `.cargo/config.toml` should no longer be needed.

dchenbecker:
@fpoli I think I made the right changes so that the top-level duchess build doesn't build the Java sources and UI tests. Please let me know if that looks better

.cargo/config.toml
```diff
@@ -1,2 +1,2 @@
 [env]
-CLASSPATH = { value = "java", relative = true, force = true }
+CLASSPATH = { value = "test-crates/duchess-java-tests/target/java", relative = true, force = true }
```
fpoli, line 2: I tried playing with this. Giving a value to `CLASSPATH` is apparently required, but the actual value is irrelevant. For example, `value = "inexistent"` would also work.

What about we remove this file, adapting the `javap` call to not pass `-cp` when `CLASSPATH` is unset?
1. In [`reflect.rs`](https://github.com/duchess-rs/duchess/blob/43c6bc5afe77f3ea4007a7ff8e86679d02b31055/macro/src/reflect.rs#L141-L151), instead of panicking we should still execute the command, but without passing `.arg("-cp").arg(classpath)`.
2. After that change, removing this `.cargo/config.toml` file should no longer break `cargo build`.

test-crates/duchess-java-tests/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+CLASSPATH = { value = "target/java", relative = true, force = true }
```
fpoli, line 2: Could you check if this file is really required? Removing it doesn't seem to cause issues when I tried it.

test-crates/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[env]
+CLASSPATH = { value = "duchess-java-tests/target/java", relative = true, force = true }
```
fpoli, line 2: Instead of adding this file, could you add a `println!("cargo:rustc-env=CLASSPATH=...");` line to the `build.rs` of `duchess-java-tests`?

Each crate in this `test-crates` currently requires its own `CLASSPATH` value, but this `.cargo/config.toml` is currently declaring a global one. The tests of `viper` are not failing just because the `CLASSPATH` declared in `viper/build.rs` seems to have precedence, but I'd prefer to not rely on that.

test-crates/duchess-java-tests/build.rs
```diff
@@ -0,0 +1,80 @@
...
+
+// These two control where this script looks for source and corresponding class files
+const SOURCE_PATH: &str = "java";
+const TARGET_PATH: &str = "target";
+
+fn main() -> std::io::Result<()> {
+    // Rerun java build if any source file changes, but then we'll check each file individually below
+    println!("cargo:rerun-if-changed={}", SOURCE_PATH);
+
+    let target_dir = Path::new(TARGET_PATH);
```
fpoli, line 16: This `test-crates/duchess-java-tests/target` folder will not be deleted by any `cargo clean`. What about placing the target dir in `../target/duchess-java-tests/`, so that it'll go away with a `cd test-crates && cargo clean`?

fpoli (commented):
Thank you! I left some comments regarding the remaining `.cargo/config.toml` files.

dchenbecker:
@fpoli thanks for the feedback! It took a little bit of experimentation but I think I was able to address everything you noted. No more `.cargo/config.toml` files and the Java classes end up in a directory that `cargo clean` will actually clean :)

test-crates/duchess-java-tests/tests/ui.rs
fpoli, line 1: I just noticed that line 18 should be `config.out_dir = Some("../target/ui".into());` to place the binaries in a folder cleaned by `cargo`.

fpoli (commented):
Great, thanks! Two more things if you don't mind. I think these are really the last changes :)

The CI is failing because the programs in the `duchess-test-java/example` folder are not built with `build.rs` (`examples` never use it). So, as a fix I'd convert those examples to ui tests, moving the folder to `duchess-test-java/tests/ui/examples`, prepending a `//@run` to all the new tests, and in particular splitting the `log.rs` example to 5 tests so that each `#[test]` in it becomes a `main` function.

I also noticed that there are 4 urls in the repository that refer to the `examples` folder (list them with `git grep /examples`). So, to avoid breaking them the links should be updated to `[...]/test-crates/duchess-java-tests/tests/ui/examples`.

dchenbecker:
Sadly there are some failing tests in examples that I missed, so let me figure that out

dchenbecker:
OK, I moved the examples to ui tests, but they all fail because we're missing stdout files for each one. At first I tried creating them, but then I ran into a problem: with the log tests, the timestamp of the log event changes on each invocation. Looking at the ui_test crate docs, it looks like the `//@run` should just check for a success code on exit. Where are we comparing stdout and stderr, and can we disable that for a subset of the tests?

fpoli:
> I've been digging through the ui_tests crate and it looks like there are some facilities for filtering/regex matching stderr, but nothing for stdout :(

Oh :( Are you referring to the fact that `//@normalize-stderr-test` is supported but `//@normalize-stdout-test` is not? If we open an issue they'll hopefully add it sooner or later. In the meantime, we might use  `//@check_pass` to check at least that the test compiles.

dchenbecker:
OK, I updated to `@check-pass` and it worked. PR is updated!

fpoli:
> I also noticed that there are 4 urls in the repository that refer to the `examples` folder (list them with `git grep /examples`). So, to avoid breaking them the links should be update to contain `test-crates/duchess-java-tests/tests/ui/example`.

Thanks! Do you also have time for this? Otherwise I'll merge and edit the links later.

.gitignore
```diff
@@ -2,3 +2,4 @@
 /Cargo.lock
 /test-crates/target
 /test-crates/Cargo.lock
+/test-crates/duchess-java-tests/target
```
fpoli, line 5: I think this is no longer needed.

```suggestion
```

dchenbecker:
Fixed the links in docs and removed the extraneous `.gitignore` entry

fpoli:
Thanks!

PR #111, opened 2023-08-07
Run tests on macOS and Windows

fpoli:
This PR tests duchess on Windows and macOS as part of the CI.

For simplicity, `mdBook test` is still only executed on Ubuntu. To have something cross-platform for mdbook it might be better to use some published GitHub action.

Note that I upgraded the `ui_test` dependency from version `0.10` to `0.15`. Since `ui_test` `0.15` uses recently stabilized Rust features, users might have to do a `rustup update` before compiling duchess.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,8 +11,11 @@ env:
 
 jobs:
   build:
-
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


fpoli:
Something is now working with the latest `ui_tests`; I keep getting "unexpected argument '--extern' found" errors.

nikomatsakis:
@fpoli maybe disable windows for now?

.github/workflows/rust.yml
```diff
@@ -11,8 +11,11 @@ env:
 
 jobs:
   build:
-
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
fpoli, line 16:
```suggestion
        # windows-latest does not work because of https://github.com/oli-obk/ui_test/issues/147
        os: [ubuntu-latest, macos-latest]
```

PR #112, opened 2023-09-24
Fix str and local free bugs

chase-ok:
This commit fixes a few safety bugs in the string and local ref code. It also cleans up the `invoke` code to make it easier to always free local refs. The `HasEnvPtr` trait was no longer providing any value so removed it along the way.

Bugs:
 1. Returning an empty string from Java would result in writing to a null pointer, since `Vec::with_capacity(0)` does not allocate and the JNI string functions always write a trailing nul byte.
 2. Allocate one extra byte in general when decoding Java strings to account for the trailing nul.
 3. Always wrap the return result of JNI function calls that can return a Java object in a `Local` before doing anything else, like checking for an exception. This ensures `Local::drop` is called and the ref is correctly freed.
 4. Fix up a few places where exceptions needed to be checked and weren't and vice versa.
 5. Fix up `?Sized` bounds so that `"abc".to_java()` compiles

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -13,7 +13,7 @@ use std::{
 
 use jni_sys::jvalue;
 
-use crate::{jvm::JavaObjectExt, Error, GlobalResult, JavaObject, Jvm, Local};
+use crate::{jvm::JavaObjectExt, Error, GlobalResult, JavaObject, Local};
 
 const VERSION: jni_sys::jint = jni_sys::JNI_VERSION_1_8;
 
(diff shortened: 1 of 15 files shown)


src/raw.rs
```diff
@@ -229,13 +229,38 @@ impl<'jvm> EnvPtr<'jvm> {
...
+    }
+
+    /// Invoke a JNI method dispatched through a virtual table lookup. Does *not* check for an exception and should
+    /// only be used when other mechanisms can prove the absence of an exception (e.g. a non-null return value or a
+    /// separate call to [`Self::check_exception()`]).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the [`jni_sys::JNIEnv`] raw pointer is only used for this invocation.
+    pub(crate) unsafe fn invoke_unchecked<F, T>(
```
nikomatsakis, line 263: Should this be private?

src/raw.rs
```diff
@@ -229,13 +229,38 @@ impl<'jvm> EnvPtr<'jvm> {
...
+    ///
+    /// First invokes [`FromJniValue::from_jni_value()`] before checking for a JVM exception. This allows [`Local`] drop
+    /// code to run and safely decrement the ref count before exiting early on an exception.
     ///
     /// # Safety
     ///
     /// The caller must ensure that the [`jni_sys::JNIEnv`] raw pointer is only used for this invocation.
     #[doc(hidden)]
-    pub unsafe fn invoke<F, T>(
+    pub unsafe fn invoke_checked<F, T: FromJniValue<'jvm>>(
```
nikomatsakis, line 242: Maybe just call this `invoke`?

nikomatsakis (changes requested):
What do you think @chase-ok ?

nikomatsakis:
Happy to merge and address further possible changes in a follow-up, as well

src/raw.rs
chase-ok: `invoke` works for me!

src/raw.rs
chase-ok: We use it in a couple of places throughout the crate (array fns, method/field lookups, string manipulation). Those places don't always return something that has a `FromJniValue` return value (e.g. method IDs or status codes) and the return value also determines whether an exception has occurred or not.

chase-ok: At least it's not `pub` anymore :D

nikomatsakis (approved):

PR #118, opened 2023-11-03
remove `pub` from package in `java_package!` example

davidhewitt:
Thank you for building duchess, I've had a lot of fun testing it out today to replace a house of cards I have in a toy project! I found that having the `pub` present in the macro input would lead to a compile error along the  lines of "unexpected input".

I suspect that at some point this support for visibility was removed? Or maybe it's a placeholder for functionality yet planned for implementation?

### DIFF
--- a/book/src/call_java_from_rust.md
+++ b/book/src/call_java_from_rust.md
@@ -28,7 +28,7 @@ Using duchess, we can declare a Rust version of this class with the `java_packag
 duchess::java_package! {
     // First, identify the package you are mirroring,
     // and the visibility level that you want.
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Hmm. I'm not sure why `pub` is not accepted -- prob just forgot to implement it!

nikomatsakis:
It appears we *always* emit public modules right now:

https://github.com/duchess-rs/duchess/blob/3346e9b8957de2393310f56b3364eaebdb2c77e2/macro/src/codegen.rs#L61

PR #124, opened 2024-03-12
support duchess::Null for passing null values

### LINKED ISSUE #38: Ability to pass `null` to methods taking nullable parameters
Sometimes we need to be able to pass a `null` parameter value when invoking a Java method. We already wrap Java method return values that might be `null` in a Rust `Option`, but wrapping in `Option` for all nullable parameters would litter user code with `Some()` in the common case. 

Is there a better way to use `null` values without the boilerplate? One thought is to have a special `struct JavaNull` that `impl<T: JavaObject> IntoJava<T>`, but having two different ways of representing null is confusing at best.

Or, could we add a new `trait IntoJavaObject` that abstracts over both `&T` (non-nullable) and `Option<&T>` (nullable) for `T: JavaObject`?

nikomatsakis:
This is much more convenient than e.g. using Option.

Fixes #38


chase-ok (approved):
The UI tests broke unrelated to this change btw

```
  -error: aborting due to previous error
  +error: aborting due to 1 previous error
```

Issue #123, opened 2024-02-28
Lifetime parameter support for ToJava


nonguyen1:
Issue reflecting structs that have references

ie.
```
//@run
use duchess::{java, prelude::*};

#[derive(duchess::ToJava)]
struct WithStringRef<'a> {
    field1: &'a str,
}

pub fn main() -> duchess::GlobalResult<()> {
    let mystring = String::from("test");
    let test = WithStringRef {
        field1: &mystring,
    };
    Ok(())
}
```
results in 

```
error: There were 1 unmatched diagnostics
 --> tests/ui/derive_scalar_fields.rs:5:8
  |
5 | struct WithStringRef<'a> {
  |        ^^^^^^^^^^^^^ Error: generic structs not yet supported
  |

full stderr:
error: generic structs not yet supported
 --> tests/ui/derive_scalar_fields.rs:5:8
  |
5 | struct WithStringRef<'a> {
  |        ^^^^^^^^^^^^^

warning: unused imports: `java`, `prelude::*`
 --> tests/ui/derive_scalar_fields.rs:2:15
  |
2 | use duchess::{java, prelude::*};
  |               ^^^^  ^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

error: aborting due to previous error; 1 warning emitted
```

nonguyen1:
https://github.com/duchess-rs/duchess/commit/aa3bc34e895e95e18a2b1e2d1cc834bcbbcb4cf7

PR #128, opened 2024-03-14
Add generic support

nonguyen1:
https://github.com/duchess-rs/duchess/issues/123

### DIFF
--- a/src/java.rs
+++ b/src/java.rs
@@ -94,6 +94,20 @@ mod auto {
             public abstract java.lang.String toString();
         }
 
+        public final class java.lang.Long {
+            public static long parseLong(java.lang.String) throws java.lang.NumberFormatException;
+            public static long parseUnsignedLong(java.lang.String) throws java.lang.NumberFormatException;
+            public static java.lang.Long valueOf(java.lang.String) throws java.lang.NumberFormatException;
+            public static java.lang.Long decode(java.lang.String) throws java.lang.NumberFormatException;
+            public java.lang.Long(java.lang.String) throws java.lang.NumberFormatException;
+            public byte byteValue();
+            public short shortValue();
+            public int intValue();
+            public long longValue();
+            public float floatValue();
+            public double doubleValue();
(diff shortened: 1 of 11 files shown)


src/java.rs
```diff
@@ -223,7 +237,7 @@ mod auto {
             //   public java.util.Date(int, int, int);
             //   public java.util.Date(int, int, int, int, int);
             //   public java.util.Date(int, int, int, int, int, int);
-            //   public java.util.Date(java.lang.String);
+            //  public java.util.Date(java.lang.String);
```
nikomatsakis, line 240:
```suggestion
            //   public java.util.Date(java.lang.String);
```

this does not look intentional

src/to_java.rs
```diff
@@ -46,7 +46,6 @@ impl<R: ?Sized> ToJava for R {
         }
     }
 }
-
 #[derive_where::derive_where(Copy, Clone)]
```
nikomatsakis, line 49:
```suggestion

#[derive_where::derive_where(Copy, Clone)]
```

nor does this :)

src/to_java.rs
```diff
@@ -206,15 +205,16 @@ where
...
 where
     J: Upcast<java::lang::Object>,
+    R: ?Sized + ToJavaImpl<J>
 {
     fn to_java_impl<'jvm>(
         rust: &Self,
         jvm: &mut Jvm<'jvm>,
     ) -> crate::Result<'jvm, Option<Local<'jvm, J>>> {
-        Ok(Some(jvm.local(rust)))
+        ToJavaImpl::to_java_impl(*rust, jvm)
```
nikomatsakis, line 217:
```suggestion
        R::to_java_impl(rust, jvm)
```

In general, my preference in these kind of situations is to specify the receiver type (`R`) that we are invoking the method on. It also means we don't have to on `*rust` but can let the compiler manage the derefs.

nikomatsakis (approved):
This looks great! Not too hard, nice. I left a few nits. I'll apply them automatically if I can.

PR #134, opened 2024-04-10
Coverage testing

rcoh:
1. Adds testing against JDK 21
2. Adds a `just coverage` target. This produces a coverage report that covers both the duchess library & the proc macro as well. It changes the ui-test framework to conditionally run rustc with `-C instrument-coverage`.
3. Adds some unit tests to validate that the coverage report is working.

In CI, a the coverage report is added as an archive.

## Why doesn't this use `cargo llvm-cov`?

Cargo llvm-cov seeks to accomplish a lot of the same things, but it can't properly (and I don't think it can be configured to) properly construct the right llvm-cov incantation to produce a unified coverage report. I wasn't able to get it to not just crash.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,6 +6,11 @@ on:
   pull_request:
     branches: [ "main" ]
 
+# Allow one instance of this workflow per pull request, and cancel older runs when new changes are pushed
+concurrency:
+  group: ci-yaml-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   CARGO_TERM_COLOR: always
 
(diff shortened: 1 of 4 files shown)


rcoh:
[ui-coverage-report.zip](https://github.com/duchess-rs/duchess/files/14936825/ui-coverage-report.zip)
Generated coverage report attached

nikomatsakis (approved):
@rcoh this looks good to me -- is it ready to merge?

nikomatsakis:
I guess, I didn't look closely at where this report is being uploaded etc. What is the plan there?

rcoh:
it's in the Checks -> Summary tab–there are much better ways to do that but we can discuss what we want. This is good to merge.

nikomatsakis:
Ah, I see. Hmm, I'll investigate, it'd be cool to run these tests as part of the mdbook building process so that we can include them and make them visible in the mdbook that goes on github pages or something.

PR #136, opened 2024-04-22
More work on threat model

nikomatsakis:
Finished write-up, added some tests


nikomatsakis:
I don't know why we are getting these errors. My hunch is that we are running with a different toolchain than I used locally. Annoying.

nikomatsakis:
cc @oli-obk any advice here?

oli-obk:
Yes, ui tests can change whenever you update your compiler. You could ignore stderr entirely and only rely on annotations in the test file if that is sufficient for your use case.

Otherwise, a rustup toolchain file is useful to ensure consistency across all contributors

nikomatsakis:
@oli-obk I think annotations would suffice, how do I make that change?

fpoli:
I think oli is referring to the `//~ ERROR` annotations that you already wrote. So, it should be enough to remove the `.stderr` files from this PR.

oli-obk:
We'll need to switch ui_test to using `OutputConflictHandling::Ignore` in the Config struct

nikomatsakis:
@oli-obk like that?

oli-obk:
Yes, this is how bevy does it, too.

If you still want it updated regularly, a cronjob could auto-open PRs if the output changes

nikomatsakis:
Oh, that'd be useful, no idea how hard that'd be to set up tho

PR #137, opened 2024-04-24
Automatically bless tests once a night

oli-obk:
implements proposal from https://github.com/duchess-rs/duchess/pull/136#issuecomment-2075658064


nikomatsakis (approved):
why not

rcoh:
we actually have a rust-toolchain file in the test-crates to make CI consistent. We can remove it now I think though?

PR #147, opened 2024-05-04
simplify public API, only allowing global JVM references

### LINKED ISSUE #142: Exceptions don't don't work properly within `Jvm::with`
In order to compute the error message for a message, duchess uses:

```rust
fn try_extract_message(exception: &impl AsJRef<Throwable>) -> String {
    let message = Jvm::with(|jvm| {
        let exception = jvm.local(exception.as_jref()?);
        exception
            .to_string()
            .assert_not_null()
            .to_rust()
            .execute_with(jvm)
    });
    message.unwrap_or_else(|err| format!("failed to get message: {}", err))
}
```

But this uses `Jvm::with` which is not re-entrant. I'm not sure the best way to work around this. Maybe we should compute the error message automatically before returning the error back to Rust? Computing the error message during `Debug` seems like it will inherently run into this issue.

Reproducer:
```rust
Jvm::with(|jvm| {
  a_java_method_that_throws_an_exception.execute_with(jvm).unwrap()
})
```

Inner error `Error::NestedUsage` even though it does not appear to be nested usage from the customer perspective.

nikomatsakis:
This makes a number of simplifications to the public API for duchess. The idea is that we focus on "global" references by default. The `duchess::Result<T>` (for example) embeds a global reference to an exception, and the new default handle is `duchess::Java<J>`, which is a global reference. The prelude has been extended to include `java` and `Java<T>` as well, for convenience.

The `to_rust`, `global`, and `execute` combinators have all been merged. You just invoke `execute` and then the result depends on the return value: returning a `Java<T>` will create a global reference (like `global` used to do), returning a Rust value will invoke the "to rust" conversion (like `to_rust` used to do).

You can no longer obtain explicit handles to the JVM, preventing panics due to nested `Jvm::with` invocations (and hence fixing #142). `Jvm::with` is in fact crate local. This also avoids some potential unsoundness results that would have arisen from nested JVM usage.

Fixes #142


book/src/threat_model.md
```diff
@@ -39,7 +39,7 @@ This section introduces invariants maintained by Duchess using Rust's type syste
 
 ### The `'jvm` lifetime `&mut Jvm<'jvm>` is the innermost scope for local variables
 
-References to Java objects of type `J` are stored in a `Local<'jvm, J>` holder. Local references can come from the arguments to native functions or from `JvmOp::execute_with` calls. `execute_with` calls use the `'jvm`' lifetime found on the `Jvm<'jvm>` argument. This allows the `Local` to be used freely within that scope. It is therefore important that `'jvm` be constrained to the **innermost** valid scope.
+References to Java objects of type `J` are stored in a `Local<'jvm, J>` holder. Local references can come from the arguments to native functions or from `JvmOp::do_jni` calls. `do_jni` calls use the `'jvm`' lifetime found on the `Jvm<'jvm>` argument. This allows the `Local` to be used freely within that scope. It is therefore important that `'jvm` be constrained to the **innermost** valid scope.
```
rcoh, line 42: I think we can delete the below section about `Jvm::with` now since `Jvm::with` is `pub(crate)`

src/jvm.rs
```diff
@@ -72,45 +74,45 @@ pub trait JvmOp: Copy {
...
     where
-        for<'jvm> Self: JvmOp<Output<'jvm> = R>,
+        for<'j> Self::Output<'j>: IntoRust<R>,
     {
-        Jvm::with(|jvm| self.execute_with(jvm))
+        ToRustOp::new(self).do_jni(jvm)
     }
-
-    fn execute_with<'jvm>(self, jvm: &mut Jvm<'jvm>) -> crate::Result<'jvm, Self::Output<'jvm>>;
+    /// Internal method
```
rcoh, line 114: a slightly more detailed comment would probably be useful for folks working on the project since this is now the fundamental "do a thing" message

macro/src/derive.rs
```diff
@@ -313,7 +310,7 @@ impl Driver<'_> {
             let pattern = variant.pat();
             return Ok(quote_spanned!(self.span() =>
                 #pattern => {
-                    #binding .jderef().upcast().execute_with(jvm)
```
rcoh, line 316: when would one call `execute_with` vs. `do_jni`? Both are on the `JvmOp` trait AFAICT and it seems like sometimes we call one or the other?

src/jvm/test.rs
```diff
@@ -1,10 +1,10 @@
-use duchess::Jvm;
+use crate::Jvm;
 
 #[test]
 fn nested_jvm_with() {
     Jvm::with(|_jvm| {
         let err = Jvm::with(|_jvm| Ok(())).expect_err("nested JVMs are illegal");
-        assert!(matches!(err, duchess::Error::NestedUsage));
+        assert!(matches!(err, crate::Error::NestedUsage));
```
rcoh, line 7: is this still a useful test since `Jvm::with` is pub_crate now? I guess for our own internal sake we want to avoid nesting it?

src/error.rs
```diff
@@ -66,7 +60,7 @@ where
 }
 
 impl<'jvm> Error<Local<'jvm, Throwable>> {
-    pub fn into_global(self, jvm: &mut Jvm<'jvm>) -> Error<Global<Throwable>> {
```
rcoh, line 69: should this still be called `into_global`? It seems we're mostly erradicating the word from the codebase

src/jvm.rs
rcoh: since `execute_with` converts into `Rust` vs. `do_jni` which keeps it in Java maybe we could rename `execute_with` something that mentioned the conversion into Rust

src/thread.rs
```diff
@@ -101,7 +101,7 @@ pub fn attach_permanently(jvm: JvmPtr) -> GlobalResult<AttachGuard> {
     })
 }
 
-pub unsafe fn attach<'jvm>(jvm: JvmPtr) -> GlobalResult<AttachGuard> {
```
rcoh, line 104: while we're here we should add a safety section to the docs or make this not pub

rcoh (approved):
looks good, had some naming comments on the internal APIs which can defer.

Probably will also need a changelog / update guidance for the next major release.

PR #151, opened 2024-05-15
Add `must_use` to JvmOp

rcoh:
It's very easy to accidentally avoid calling execute because of how Rust-like Duchess makes calling Java code. As an example, one of our own examples failed to call execute!


nikomatsakis (approved):

nikomatsakis:
Good call.

PR #157, opened 2024-05-21
Add test for null String

nonguyen1:
Add a simple test for java `String` is `null` from Rust None `Option`.

### DIFF
--- a/test-crates/duchess-java-tests/tests/ui/passing_null_values.rs
+++ b/test-crates/duchess-java-tests/tests/ui/passing_null_values.rs
@@ -21,5 +21,10 @@ pub fn main() -> duchess::Result<()> {
     let is_null = take_null.take_null_string(duchess::Null).execute()?;
     assert!(is_null);
 
+    let is_null = take_null
+        .take_null_string(&None::<Java<java::lang::String>>)
+        .execute()?;
+    assert!(is_null);
+
     Ok(())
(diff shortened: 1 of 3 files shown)


test-crates/duchess-java-tests/tests/ui/passing_null_values.rs
```diff
@@ -5,6 +5,13 @@ duchess::java_package! {
     package take_null;
 
     public class TakeNull { * }
+    public class TakeNullRecord { * }
```
rcoh, line 8: we need to change this test to `//@run` — currently it's not actually being run, just compiled.

rcoh (approved):
can you change the test to `//@run`?

nonguyen1:
We'll need to hold off on this. After I updated the test to `//@run` I'm seeing some runtime failures. I'll try to take a look soon.

```rust
    let java = take_null_record.to_java().assert_not_null().execute()?;
    let is_null = java.is_null().execute().unwrap(); // <- `Error: attempted to deref a null Java object pointer`
    assert!(!is_null);
```

nikomatsakis:
So is this fixed by https://github.com/duchess-rs/duchess/pull/182 ?

nonguyen1:
Yup fixed by https://github.com/duchess-rs/duchess/pull/182

PR #159, opened 2024-05-21
Some bless/autobless changes

oli-obk:
The autobless still won't work, as it doesn't have permissions, but at least it should do minimal work now

### DIFF
--- a/test-crates/duchess-java-tests/tests/ui/duplicate_implements.stderr
+++ b/test-crates/duchess-java-tests/tests/ui/duplicate_implements.stderr
@@ -1,5 +1,5 @@
 error: error in class `log.Builder`: duplicate reference in 'implements' to 'log.BuildStep'
- --> $DIR/duplicate_implements.rs:4:5
+ --> tests/ui/duplicate_implements.rs:4:5
   |
 4 |     public class log.Builder implements log.BuildStep, log.BuildStep {}
   |     ^^^^^^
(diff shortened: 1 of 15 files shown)


nikomatsakis:
@oli-obk you have perms now to adjust things, so shall we assume you're just going to tweak this PR until it works?

test-crates/duchess-java-tests/tests/ui/duplicate_implements.stderr
```diff
@@ -1,5 +1,5 @@
 error: error in class `log.Builder`: duplicate reference in 'implements' to 'log.BuildStep'
- --> $DIR/duplicate_implements.rs:4:5
```
rcoh, line 2: it's ok to replace `$DIR` with `test/ui` here?

rcoh (commented):
Can we add a README for why we have these files? I would definitely otherwise assume that they are load bearing, but it seems like they're just information?

test-crates/duchess-java-tests/tests/ui/duplicate_implements.stderr
oli-obk: yea, this is a leftover from the last ui_test version bump

rcoh (approved):
lgtm, we can also switch to using a rust-toolchain.toml, maybe the better long term solution

Issue #158, opened 2024-05-21
Fix asserts in UI tests


nonguyen1:
Asserts under under `duchess-java-tests/tests/ui` do not validate values.

## Reproduce

1. Checkout pkg
2. Add `assert!(false)` to any test.
3. from `duchess-java-tests` run `cargo test`

All tests should still pass.

rcoh:
yeah, this is because of the (surprising) behavior that most of these are `//@check-pass` and not `//@run`

rcoh:
if you're making CRs in this area, changing the tests that are `//@check-pass` to `//@run` would be welcome.

nonguyen1:
Hehe this makes sense now haha. Sure I can update them.

PR #160, opened 2024-05-21
Run existing UI tests

nonguyen1:
Update all `ui` tests from`@check-pass` to `@run`.

Ignores `ui/passing_null_values.rs` update bc it fails for `Error: attempted to deref a null Java object pointer`

### DIFF
--- a/test-crates/duchess-java-tests/tests/ui/examples/auth.rs
+++ b/test-crates/duchess-java-tests/tests/ui/examples/auth.rs
@@ -1,4 +1,4 @@
-//@check-pass
+//@run
 use duchess::prelude::*;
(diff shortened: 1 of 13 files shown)


nonguyen1:
Related https://github.com/duchess-rs/duchess/issues/158

test-crates/duchess-java-tests/tests/ui/examples/auth.rs
```diff
@@ -1,4 +1,4 @@
-//@check-pass
+//@run
```
oli-obk, line 1: Once all tests are in `run` mode, we can switch them globally and remove all the annotations

PR #162, opened 2024-05-21
Update thread model to include new understandings

rcoh:
After some investigation, there does not seem to be any loss of safety possible when incorrect `.class` files are used.

### DIFF
--- a/book/src/safety.md
+++ b/book/src/safety.md
@@ -11,7 +11,7 @@ However, there are edge cases that can "void" this guarantee and which Duchess c
 
 Duchess will guarantee memory safety within your crate, but there are two conditions that it cannot by itself guarantee:
 * Duchess does not "sandbox" or "make safe" the Java code it executes. You MUST ensure that Java code being invoked is safe and trusted. You MUST NOT invoke untrusted Java code with Duchess.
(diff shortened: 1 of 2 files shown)


book/src/safety.md
```diff
@@ -10,10 +10,9 @@ However, there are edge cases that can "void" this guarantee and which Duchess c
 ## Memory safety requirements
 
 Duchess will guarantee memory safety within your crate, but there are two conditions that it cannot by itself guarantee:
-
-* **You must build with the same Java class files that you will use when you deploy:**
-    * Part of how Duchess guarantees is safety is by reflecting on `.class` files at build time.
-    * If you build against one set of class files then deploy with another, 
+* Duchess does not "sandbox" or "make safe" the Java code it executes. You MUST ensure that Java code being invoked is safe and trusted. You MUST NOT invoke untrusted Java code with Duchess.
+* **You must SHOULD with the same Java class files that you will use when you deploy:**
```
nikomatsakis, line 14:
```suggestion
* **You SHOULD with the same Java class files that you will use when you deploy:**
```

nikomatsakis (approved):

PR #163, opened 2024-05-31
Add JMX APIs to Java prelude

rcoh:
Currently, there is not straightforward way of defining a `java.*` package outside of Duchess

### DIFF
--- a/src/java.rs
+++ b/src/java.rs
@@ -343,6 +343,7 @@ mod auto {
 
         public class ManagementFactory {
             public static java.util.List<java.lang.management.GarbageCollectorMXBean> getGarbageCollectorMXBeans();
+            public static java.util.List<java.lang.management.MemoryPoolMXBean> getMemoryPoolMXBeans();
             public static java.lang.management.MemoryMXBean getMemoryMXBean();
         }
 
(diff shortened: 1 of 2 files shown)


wleslie:
In general, how do we decide which Java APIs go into the Duchess prelude? So far we've got a rather arbitrary subset of the Java standard library. If we continued adding whichever APIs any user needs, would we eventually end up reflecting the entire Java standard library here?

PatelProjects:
The `GarbageCollectorMXBean` part looks good

I'd like the `MemoryPoolMXBean` interface exposing at least `getType()`, `getUsage` and `getCollectionUsage`, instead of the `MemoryMXBean`

nikomatsakis:
A thought (not for this PR) is to have duchess export java.lang and java.util (or some set of known packages) and then have ability for other crates to do other stuff. Then we can even have duchess-java-foo crates too for convenience.

wleslie:
@rcoh Makes sense. Assuming that we provide a general way to reflect arbitrary `java.*` classes in the future, would we be OK doing a breaking change to remove the JMX stuff from the prelude?

rcoh:
maybe we will feature gate them or move them to a separate crate. However, I'd imagine for many folks access to the JMX APIs is going to be a generally useful feature.

PR #168, opened 2024-06-25
Add in checks for the type of Java Class/Interface

### LINKED ISSUE #67: distinguish class vs interface
We currently call everything a 'class' in `duchess::java_package@!`. We should distinguish class vs interface and enforce it in the `check` macro.

ekureina:
Works for both fully reflected classes and for specified classes. Added UI tests for both cases, and updated other portions where needed to specify the correct type of Java import.

For Reflected classes, fixed an error in specifying the span and kind of the class. Previously, they took the span of the reflection cache, which meant that the error could confusingly point to the main `java_package!` macro invocation.

Fixes #67.

### DIFF
--- a/test-crates/duchess-java-tests/tests/ui/interface_class_mismatch_specified.stderr
+++ b/test-crates/duchess-java-tests/tests/ui/interface_class_mismatch_specified.stderr
@@ -1,0 +1,8 @@
+error: error in class `exceptions.ThrowExceptions`: class `exceptions.ThrowExceptions` should be type `Class`
+ --> tests/ui/interface_class_mismatch_specified.rs:4:5
+  |
+4 |     public interface ThrowExceptions { }
+  |     ^^^^^^
+
+error: aborting due to 1 previous error
+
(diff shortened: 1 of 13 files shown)


test-crates/duchess-java-tests/tests/ui/interface_class_mismatch_specified.stderr
```diff
@@ -0,0 +1,8 @@
+error: error in class `exceptions.ThrowExceptions`: class `exceptions.ThrowExceptions` should be type `Class`
```
rcoh, line 1: two small nits on this error:
1. the span is wrong—it should be on `interface` and not `public`
2. Should we lower case the output so that it's actually correct?

rcoh (commented):
looks good. We'll need to make another major release—this would be a breaking change for anyone that happened to get it wrong.

test-crates/duchess-java-tests/tests/ui/interface_class_mismatch_specified.stderr
ekureina: 1. Yes, but that's where all the Spans are, even for other things. I'd prefer to punt that to a separate issue where we try and have a better place for the spans.
2. Yeah, that's probably better.

ekureina:
Yep, didn't expect it to not be a breaking change release.

nikomatsakis (approved):

nikomatsakis:
Nice! Once UI checks pass, let's merge.

Issue #166, opened 2024-06-24
java_function not handling packages with underscores correctly


bowemi:
## Summary

In the JNI specification, underscores are escaped with "_1". Duchess does not take this into account when generated the JNI functions.

https://docs.oracle.com/en/java/javase/17/docs/specs/jni/design.html

## Example

Given this java class

```
package com.example.michael_test_client;

class TestClient {
    public static native String test();

    public static void main(String[] args) {
        System.load("/home/bowemi/workplace/duchess_bug/target/release/libmichael_test.so");
        System.out.println(test());
    }
}
```

and this lib.rs

```
use duchess::{java, prelude::*};
use jni::sys::jstring;
use jni::JNIEnv;
use jni::objects::JClass;


duchess::java_package! {
    package com.example.michael_test_client;

    class TestClient { * }
}

#[duchess::java_function(com.example.michael_test_client.TestClient::test)]
fn test_function<'n>() -> duchess::LocalResult<'n, String> {
    Ok("duchess".to_string())
}

#[no_mangle]
pub extern "system" fn Java_com_example_michael_1test_1client_TestClient_test<
    'local,
>(
    mut env: JNIEnv<'local>,
    _class: JClass<'local>,
) -> jstring {
    let output = env.new_string("jni").unwrap();
    output.into_raw()
}
```

"jni" gets printed because the name is correctly defined, whereas duchess generates the name ```Java_com_example_michael_test_client_TestClient_test```

PR #169, opened 2024-06-25
Generate JNI fuction following JNI standards

bowemi:
When java package names contain `_`, the underscore needs to be escaped
as `_1`. Without this escaping, Java can not find the native function

This change also proposes a test framework for Java initiated
applications calling into Rust generated functions.

fixes https://github.com/duchess-rs/duchess/issues/166


macro/src/class_info.rs
```diff
@@ -708,10 +708,27 @@ impl DotId {
...
+        self.split().1.data.replace("_", "_1").into()
+    }
+
     pub fn class_name(&self) -> &Id {
         self.split().1
     }
 
+    /// returns the package in JNI format with _'s escaped with _1
+    /// https://docs.oracle.com/en/java/javase/17/docs/specs/jni/design.html
+    pub fn to_jni_packge(&self) -> String {
```
nikomatsakis, line 723:
```suggestion
    pub fn to_jni_package(&self) -> String {
```

test-crates/duchess-java-tests/tests/ui/native_fn_callable_from_java.rs
```diff
@@ -0,0 +1,21 @@
+//@compile-flags: --crate-type cdylib
+//@check-pass
+//@rustc-env: CLASSPATH=../target/tests/java_ui
```
nikomatsakis, line 3: Hmm, I like this capability, I don't know if I like mixing it in with the other `ui` tests. What about making a separate directory where these compile-flags and rustc-env options are the default? Something like:

* `src/test/rs-to-java`
* `src/test/rs-from-java` <-- the new one
* `src/test/js-to-rs`

or even just put the `.rs` files into the `src/test/js-to-rs` directory as well?

nikomatsakis (changes requested):
What do you think?

test-crates/duchess-java-tests/tests/ui/native_fn_callable_from_java.rs
bowemi: I like your 3 directory proposal, I'll go with that, but as far as naming and structure, what about...

* ```src/test/java-to-rust/java```
* ```src/test/java-to-rust/rust_libraries```
* ```src/test/rs-to-java```

bowemi (commented):
Moved the ui test files into the directory structure based on Niko's comment, also correct the function ```to_jni_packge``` -> ```to_jni_package```

nikomatsakis (approved):
Very nice.

nikomatsakis:
> This branch has conflicts that must be resolved

unfortunately, can't merge yet :/

Issue #173, opened 2024-07-11
Native functions propagating Java exceptions


nikomatsakis:
Currently when native functions can fail in two ways, both unimplemented:

1. Encountering a duchess error ([panics here](https://github.com/duchess-rs/duchess/blob/65b499878fd9b581e0aeb86bd51444dd1f339302/src/jvm.rs#L164))
2. Catching a Rust panic ([panics here](https://github.com/duchess-rs/duchess/blob/65b499878fd9b581e0aeb86bd51444dd1f339302/src/jvm.rs#L225))

In both cases we should be invoking the [jni throw method](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Throw) in addition to returning a sentinel failure value (null for objects, 0 for scalars). The precise value to be thrown will depend on the kind of failure

* For the case of [Error::Thrown](https://github.com/duchess-rs/duchess/blob/65b499878fd9b581e0aeb86bd51444dd1f339302/src/error.rs#L23), we should take the java object and invoke `Throw` on it.
* For [`Error::NullDeref`](https://github.com/duchess-rs/duchess/blob/65b499878fd9b581e0aeb86bd51444dd1f339302/src/error.rs#L31), we would ideally create a `NullPointerException`.
* For the other cases (including arbitrary Rust panics), we can just create a generic `RuntimeException` with an appropriate message.

Simple test case on this branch:

https://github.com/nikomatsakis/duchess-ndm/tree/java-to-rust-exceptions

nikomatsakis:
Closed by #176

PR #176, opened 2024-07-12
Java to rust exceptions

bowemi:
This change updates native_function_returning_object and native_function_returning_scalar to translate crate::Error and rust panics into a corresponding Java exception.

https://github.com/duchess-rs/duchess/issues/173

### DIFF
--- a/src/jvm.rs
+++ b/src/jvm.rs
@@ -141,20 +141,17 @@ fn throw_java_runtime_exception(env: EnvPtr<'_>, message: &str) {
     let runtime_exception_clazz = crate::java::lang::RuntimeException::class(&mut jvm)
         .expect("java/lang/RuntimeException not found");
 
+    let runtime_exception_clazz_ptr = runtime_exception_clazz.as_raw().as_ptr();
+
     let encoded = cesu8::to_java_cesu8(message);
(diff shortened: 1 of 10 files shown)


src/jvm.rs
```diff
@@ -135,6 +135,56 @@ fn get_or_default_init_jvm() -> crate::Result<JvmPtr> {
...
+
+    unsafe {
+        env.invoke_unchecked(
+            |env| env.ThrowNew,
+            |jni, f| f(jni, runtime_exception_clazz_ptr, c_string_ptr),
+        );
+    };
+}
+
+fn error_to_java_exception(env: EnvPtr<'_>, error: Error<Local<Throwable>>) {
```
nikomatsakis, line 159:
```suggestion
fn error_to_java_exception(env: EnvPtr<'_>, error: Error<Local<'_ Throwable>>) {
```

I prefer to make these lifetimes explicit, allowing me to distinguish borrowed from fully owned types at a glance.

nikomatsakis (changes requested):
Let's make a test that panics from within the Rust code as well.

nikomatsakis (approved):

PR #177, opened 2024-07-15
cherry-pick some refactorings

nikomatsakis:
I had a branch implementing the ability for Rust types to implement Java callbacks. I'm taking a fresh approach but these commits are some small refactorings that I want to land first and which will be used by the new code.

r? @rcoh

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -9,6 +9,7 @@
     - [Threat model](./threat_model.md)
 - [Reference](./reference.md)
     - [The `java_package` macro](./java_package.md)
+    - [Translating Java method signatures to Rust](./java_signatures_in_rust.md)
     - [The `java_function` macro](./java_function.md)
         - [Linking native functions into the JVM](./linking_native_functions.md)
     - [Deriving Java/Rust conversions](./derive.md)
--- a/book/src/java_signatures_in_rust.md
+++ b/book/src/java_signatures_in_rust.md
@@ -1,0 +1,26 @@
+# Translating Java method signatures to Rust
+
+The [`java_package`](./java_package.md) macro translates Java methods into Rust methods.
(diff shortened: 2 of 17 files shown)


rcoh:
trying to grok the JvmRefOp—my understanding is:

Previously, we only had (by analogy), `Future`. But we would also like `IntoFuture` because that's much more convenient to implement. Which on its own wouldn't require the creation of JvmRefOp — but `IntoJava` only makes sense for references not, scalars?

rcoh (approved):

nikomatsakis:
I should document the traits -- and we should consider if they need a new name.

Here is the motivation, for the record:

Before we accepted `IntoJava<J>` which was an alias for a `JvmOp` that produces a Java object of type `J` (technically, something that could be deref'd to `&J`, essentially, so that covers a `Local<J>` and a `Java<J>` etc). But to support callbacks, I want to be able to take a Rust value and then do some preprocessing before invoking the jni method. But this preprocessing should take place with a jvm in scope. So I changed it so that `IntoJava`, as you surmised, has a role more like `IntoFuture` -- it is something that can be made into a `JvmOp`, but not necessarily a `JvmOp` itself. And then I added a trait alias `JvmRefOp<J>` for a `JvmOp` that produces an object of type `J`.

What I had planned (and still do plan) is to be able to (1) take a Rust value that implements a Java trait; (2) box it; (3) create a java object that owns that boxed pointer (i.e., we mem::forget the box); (4) when that java object is GC'd, we'll call back into the Rust crate to free the box. 

I don't 100% remember why this couldn't be done by making the Rust type into a `JvmOp`, maybe it could but it seemed bad for other reasons?

nikomatsakis:
I'm going to land this set of changes, but let's revisit the traits more holistically.

PR #180, opened 2024-07-17
Allow to specify minimum or maximum JNI versions

ekureina:
Currently, can pick between JNI 1.6 or 1.8. Resolves to one JNI version.

Will fail to compile if users ask for conflicting JNI versions, but will try and resolve to a JNI version that everyone in the compile tree is happy with. Also updated the github workflows to run against JNI 1.8 and JNI 1.6. They are done in sets of `at_most` and `at_least`, as rust features are meant to be additive.

Will fail to compile if `jni_at_most_1_6` and `jni_at_least_1_8` is requested, which is normal and expected.


ekureina:
Do we want to have separate runs in the matrix for each JNI version?

ekureina:
> Do we want to have separate runs in the matrix for each JNI version?

Per Zulip discussion, yes. Also, we should simplify this down considerably.

nikomatsakis (approved):

PR #185, opened 2024-09-30
extract codegen into macro-rules macros

nikomatsakis:
This applies a pattern that I've found to be very nice to work with, in which the duchess crate has two helper crates:

* `duchess-macro-rules`, internal `macro_rules!` macro that handle most of the codegen
* `duchess-macros`, procedural macros which generate calls to those macros after doing validation and pre-processing

This combo lets each kind of macro manage the things it is good at: procedural macros for generating errors and doing some grody stuff, but macro-rules for generating the actual code.

Also, as a driveby change, adds a `dev-container.json` file which makes it easier to manage the setup for testing and developing.

r? @rcoh


macro-rules/src/java_types.rs
```diff
@@ -0,0 +1,269 @@
...
+#[macro_export]
+macro_rules! argument_impl_trait {
+    ($scalar:ident $(+ $lt:lifetime)?) => {
+        impl duchess::IntoScalar< duchess::plumbing::rust_ty!($scalar) > $(+ $lt)?
+    };
+
+    ($r:tt $(+ $lt:lifetime)?) => {
+        impl duchess::IntoJava< duchess::plumbing::rust_ty!($r) > $(+ $lt)?
+    };
+}
```
rcoh, line 25: for the macros that aren't single-pattern and pretty obvious, I think some usage examples would probably be helpful

macro-rules/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+mod java_types;
```
rcoh, line 1: as a small point, it's hard to find where the macros are actually defined because they all get exported into the root namespace. I guess there's no way around that though :-/

macro-rules/src/util.rs
```diff
@@ -0,0 +1,25 @@
+#[macro_export]
+macro_rules! macro_if {
+    (if [] { $($then:tt)* }) => {};
+
+    (if false { $($then:tt)* }) => {};
+
+    (if [$($input:tt)+] { $($then:tt)* }) => {
+        $($then)*
+    };
```
rcoh, line 9: I'm kinda confused, but this looks like it ignores the $input arguments? AFAICT it expands identically to `if true`

macro-rules/src/util.rs
```diff
@@ -0,0 +1,25 @@
+#[macro_export]
```
rcoh, line 1: some usage examples would be very helpful here

duchess-reflect/src/codegen.rs
```diff
@@ -134,238 +129,30 @@ impl ClassInfo {
...
-
-            #[allow(non_camel_case_types)]
-            pub struct #struct_name<#(#java_class_generics_with_defaults,)*> {
-                _empty: std::convert::Infallible,
-                _dummy: ::core::marker::PhantomData<(#(#java_class_generics,)*)>
+        let mro_tys = self.mro(upcasts)?;
+
+        let output = quote! {
+            duchess::plumbing::setup_class! {
+                struct_name: [#struct_name],
```
rcoh, line 136: just curious—why put all of them in `[]`? Does it simplify the resolution of the macro?

macro-rules/src/util.rs
```diff
@@ -0,0 +1,25 @@
...
+        $($then)*
+    };
+
+    // Testing what kind of type we have (scalar vs ref)
+    //
+    // see java_types.rs
+
+    (if is_ref_ty($t:ident) { $($then:tt)* }) => {};
+
+    (if is_ref_ty($t:tt) { $($then:tt)* }) => {
```
rcoh, line 21: this works because idents are never references but token trees are?

src/lib.rs
```diff
@@ -68,6 +68,12 @@ pub mod plumbing {
     pub use crate::raw::{EnvPtr, FieldPtr, FromJniValue, IntoJniValue, MethodPtr, ObjectPtr};
     pub use crate::refs::NullJRef;
     pub use crate::to_java::{ToJavaImpl, ToJavaScalar};
+    pub use duchess_macro_rules::{
```
rcoh, line 71: I wonder if we should make the whole plumbing module doc(hidden)

rcoh (approved):
This is definitely a big improvement in terms of readability of the codegen.

In the long term, having some tests of the macros directly is probably a reasonable thing to add—I think it will make working on the macros easier.

There is some dead (and I think? incorrect) code in `macro_if` that we should remove before merging. But otherwise this looks good.

macro-rules/src/util.rs
rcoh: ah—I think actually all of these expect the last two are dead code

nikomatsakis:
I'm of mixed minds about unit tests on the macros themselves — they don't represent a stable API boundary and I'm generally not in favor of tests that don't target a stable API boundary, since they can become dated so easily when refactorings occur. On the other hand, I know they can have value in making it easy to specify edge cases.

src/lib.rs
nikomatsakis: Maybe. I've been thinking we should also rename it to `semver_unstable` or something more explicit.

macro-rules/src/lib.rs
nikomatsakis: No, no way around that, but we could break them into a file per macro.

duchess-reflect/src/codegen.rs
nikomatsakis: It's not always necessary, but it is needed in some cases where a `,` is not a suitable "end delimiter", e.g. `$($t:tt)*` but also on occasion some other fragment specifiers. I used to only use it sometimes but I've wound up just adopting a consistent style of always including them.

macro-rules/src/util.rs
nikomatsakis: Yes, it's part of the invariant for how we represent types -- an identifier is a scalar or void, a `()` grouping is a reference. I could replace `$t:tt` with `($(t:tt)*)`, that might be better.

nikomatsakis:
Thanks for the review, @rcoh, I'll go over the docs and make a few improvements and then merge.

nikomatsakis:
Made cosmetic improvements. Will merge once CI is green.

PR #186, opened 2024-10-01
rename `s/plumbing/semver_unstable`, hide in docs

nikomatsakis:
Easy review

r? @rcoh


rcoh:
the one thing that I'm worried about here is that currently the macros package is pinned to the same version as everything else. Naming it this makes it sounds like you could make breaking changes to it without releasing a new major version of duchess (but you can't in practice).

We actually _can't_ make a breaking change in the macro package today because an older version of duchess would pick up the new macro package.

It might be worth taking care of that versioning before we accidentally cause a semver break later.

rcoh (approved):

nikomatsakis:
Ah. Yes, the macro crates should just be versioned with an independent integer (1.0, 2.0, 3.0, etc).

PR #187, opened 2024-10-22
introduce a build-rs and a `cargo-duchess` utility

nikomatsakis:
These don't really do much and they are not yet documented, but the plan is...

* the `build.rs` will run `javap` and serialize the class information for later use by the proc macro
* it can also do other preprocessing like generating java classes etc
* the proc macro will generate the code
* the CLI tool will install the dependencies and also offer some nice things like creating a JAR file


nikomatsakis:
There's *something* about that test that has a kind of race condition, but I don't know what it is...

nikomatsakis:
...it'd be nice if the build-rs could do the building of java...

duchess-build-rs/src/shim_writer.rs
```diff
@@ -0,0 +1,102 @@
...
+    java_interface_info: &'w JavapClassInfo,
+}
+
+impl<'w> ShimWriter<'w> {
+    pub fn new(
+        writer: &'w mut impl Write,
+        shim_name: &'w str,
+        java_interface_info: &'w JavapClassInfo,
+    ) -> Self {
+        ShimWriter {
```
rcoh, line 19: Do you think we should merge this as is or stash it in a branch until we are actually ready to use it?

rcoh (approved):
LGTM. My only question is if we want to prune out the things that will not be immediately used before merging

rcoh (changes requested):
~I'm not sure how this worked on Windows before, but it doesn't work on Windows anymore. I'm guessing the new classpath code isn't assembling the classpath in a windows-compatible way~

I'm pretty sure I found it, pushed a fix 🤞🏻

nikomatsakis:
oh dear god.

duchess-build-rs/src/shim_writer.rs
nikomatsakis: I am 99.9% sure I am going to want it :) I say just merge it.

rcoh (approved):

PR #193, opened 2025-01-23
Rework Java function support

nikomatsakis:
The big user-facing change here is that the arguments to ` #[java_function]` must match the Java types exactly (i.e., be `Option<&java::lang::String>` or whatever).


bowemi (commented):
https://github.com/duchess-rs/duchess/blob/main/test-crates/duchess-java-tests/tests/java-to-rust/java/java_to_rust_greeting/Java_Can_Call_Rust_Java_Function.java

Can you update or add a similar test to this one where Java calls a rust function with null

bowemi (approved):

macro/src/java_function.rs
```diff
@@ -1,7 +1,7 @@
 use std::{iter::once, sync::Arc};
 
 use duchess_reflect::{class_info::ClassInfoAccessors, reflect::PrecomputedReflector};
-use proc_macro2::{Ident, Literal, TokenStream};
+use proc_macro2::{ Literal, TokenStream};
```
rcoh, line 4: note: looks like cargo fmt was not ron

macro-rules/src/setup_java_function.rs
```diff
@@ -0,0 +1,100 @@
...
+        // The type of `this` we expect to come from Java.
+        // For static methods, this is `duchess::semver_unstable::jni_sys::jclass`.
+        // For instance methods, this will be the type of the Java object,
+        // e.g., `&java::lang::String`.
+        abi_this_ty: $abi_this_ty:ty,
+
+        // Name we will use for the "this" argument given from Java (of type `$abi_this_ty`).
+        abi_this_name: $abi_this_name:ident,
+
+        // Argument we will supply as "this" tot he Rust code (if any).
```
rcoh, line 24:
```suggestion
        // Argument we will supply as "this" to the Rust code (if any).
```

macro-rules/src/setup_java_function.rs
```diff
@@ -0,0 +1,100 @@
...
+                $($abi_argument_names: $abi_argument_tys,)*
+            ) -> $abi_return_ty {
+                // Covers the calls to the two `duchess::plumbing` functions,
+                // both of which assume they are being invoked from within a JNI
+                // method invoked by JVM. This function is anonymous and not
+                // callable otherwise (presuming user doesn't directly invoke it
+                // thanks to the `#[no_mangle]` attribute, in which case I'd say they are
+                // asking for a problem).
+                //
+                // **NB.** It's important that #rust_invocation does not contain any user-given
```
rcoh, line 66: is `#input_fn_name` `#rust_invocation`? I don't see anything specifically named `#rust_invocation`

rcoh: I assume its not that since `input_fn_name` _is_ provided by the user.

rcoh (approved):
I had some questions on comments. Overall this looks good.

I wonder if in the long term, duchess should check nullness of these values to avoidthe boilerplate for customers (or perhaps gluegun will, I guess.)

nikomatsakis:
> I wonder if in the long term, duchess should check nullness of these values to avoidthe boilerplate for customers (or perhaps gluegun will, I guess.)

I think we could do some adaptation but this seems like a good place to start

PR #197, opened 2025-05-01
Fix issue on aarch64 with jdk-17.0.15+6-LTS

Benjscho:
There seems to be some issue with launch_or_use_existing where it will panic in the try_launch branch reliably on aarch64 linux hosts (reproduced on AL2 & AL2023). Adding this block in fixes it.

This is the minimal example to reproduce the crash when using Open JDK jdk-17.0.15+6-LTS:
```rust
use jni::JavaVM;

fn main() {
    let args = jni::InitArgsBuilder::new()
        .build()
        .expect("Failed to build JVM InitArgs");
    let jvm = JavaVM::new(args).expect("Failed to build JVM");
    duchess::Jvm::builder().launch_or_use_existing().unwrap();
}
```

This patch reliably fixes the panic, not sure why right now, and I'm not enough of a JVM expert to figure out why. 

I also have this diff prepared against 0.3.0 and 0.2.1 tags to fix those versions, would appreciate new patch releases for those versions if possible.


Benjscho:
Thanks to @bowemi for figuring out the patch!

bowemi:
Can you add a test in this directory that validates this use case?

https://github.com/duchess-rs/duchess/tree/main/test-crates/duchess-java-tests/tests/rust-to-java

Basically the minimal repro you came up with...

```
use jni::JavaVM;

fn main() {
    let args = jni::InitArgsBuilder::new()
        .build()
        .expect("Failed to build JVM InitArgs");
    let jvm = JavaVM::new(args).expect("Failed to build JVM");
    duchess::Jvm::builder().launch_or_use_existing().unwrap();
}
```

rcoh:
Can we reproduce this failure in CI? Presumably we need to test that specific JVM version on AArchy64? I'd have a lot more confidence that we actually fixed it if that was the case.

rcoh:
I have created release-0.3.x and release-0.2.x for backports

bowemi (approved):
[use_existing_works.rs](https://github.com/duchess-rs/duchess/pull/197/files#diff-9412d24a9606384ff48fb396a15aa36d9f4de5782323c2625bffe38b75e7d50d)

[Nitpick] I would prefer a name that more accurately describes what the test is doing. I propose "duchess_can_use_jni_launched_jvm.rs"

bowemi:
Also validated in this PR your test captures the error without the change

https://github.com/duchess-rs/duchess/actions/runs/14799792404/job/41555631489?pr=200

bowemi (approved):

rcoh (approved):

PR #204, opened 2025-06-15
Update java_function to allow Options with &mut.

bowemi:
Also update primitive arrays to support a new function `set_array_region`. `set_array_region` calls Set<Primitive>ArrayRegion. This allows for mutability of java arrays in place without needing to transform them into rust arrays first

### DIFF
--- a/macro/src/java_function.rs
+++ b/macro/src/java_function.rs
@@ -1,9 +1,10 @@
+use core::panic;
 use std::{iter::once, sync::Arc};
 
 use duchess_reflect::{class_info::ClassInfoAccessors, reflect::PrecomputedReflector};
 use proc_macro2::{Literal, TokenStream};
 use quote::quote_spanned;
-use syn::spanned::Spanned;
+use syn::{spanned::Spanned, FnArg};
 
 use crate::{
     argument::MethodSelector,
@@ -217,6 +218,29 @@ impl Driver<'_> {
         .forbid_capture(|sig| sig.java_ty_rs(ty))?)
     }
 
+    // fn is_option_mut_ref(pat_type: &syn::PatType) -> bool {
+    fn is_option_mut_ref(fnArg: &FnArg) -> bool {
+        let pat_type = match fnArg {
+            syn::FnArg::Typed(pat_type) => pat_type,
+            syn::FnArg::Receiver(_) => return false
+        };
+
+        if let syn::Type::Path(type_path) = &*pat_type.ty {
+            if let Some(segment) = type_path.path.segments.last() {
+                if segment.ident == "Option" {
+                    if let syn::PathArguments::AngleBracketed(args) = &segment.arguments {
+                        for arg in args.args.iter() {
+                            if let syn::GenericArgument::Type(syn::Type::Reference(type_ref)) = arg {
+                                return type_ref.mutability.is_some();
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        false
+    }
+
     fn user_arguments(&self, input: &syn::ItemFn) -> syn::Result<Vec<Argument>> {
         let selector_span = self.selector.span();
         let mut arguments = vec![];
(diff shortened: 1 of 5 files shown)


macro/src/java_function.rs
```diff
@@ -217,6 +218,29 @@ impl Driver<'_> {
         .forbid_capture(|sig| sig.java_ty_rs(ty))?)
     }
 
+    // fn is_option_mut_ref(pat_type: &syn::PatType) -> bool {
+    fn is_option_mut_ref(fnArg: &FnArg) -> bool {
```
bowemi, line 222: I thought it would be nice to allow JNI functions to specify they wanted to take a mutable reference to the java objects with something like...

```
Option<&mut Array<i8>>
```
But in practice it wasn't actually useful since everything takes in ```self``` instead of ```&self```. I'd like to get thoughts on if it's worth trying to enforce mutability in this instance

src/array.rs
```diff
@@ -207,6 +224,47 @@ macro_rules! primitive_array {
...
+                This: JvmOp,
+                for<'jvm> This::Output<'jvm>: AsJRef<JavaArray<$rust>>,
+            {
+                type Output<'jvm> = ();
+            
+                fn do_jni<'jvm>(self, jvm: &mut Jvm<'jvm>) -> crate::LocalResult<'jvm, Self::Output<'jvm>> {
+                    let this = self.this.do_jni(jvm)?;
+                    let this = this.as_jref()?.as_raw();
+                
+                    unsafe {
```
rcoh, line 255: for the unsafe block, please add a safety comment talking about the invariants that are enforced and why this is safe

src/array.rs
```diff
@@ -207,6 +224,47 @@ macro_rules! primitive_array {
...
                 }
             }
+
+            impl<This> JavaArrayModificationExt<JavaArray<$rust>, $rust> for This
+            where
+                This: JvmOp,
+                for<'jvm> This::Output<'jvm>: AsJRef<JavaArray<$rust>>,
+            {
+                type SetArrayRegion<'a> = SetArrayRegion<'a, Self, JavaArray<$rust>, $rust>;
+                fn set_array_region<'a>(mut self, start: usize, values: &'a[$rust]) -> Self::SetArrayRegion<'a> {
```
rcoh, line 234: are there Rust std library APIs that match this pattern of offset + slice?

src/array.rs
```diff
@@ -86,6 +89,11 @@ pub trait JavaArrayExt<T: JavaType>: JvmOp {
     fn length(self) -> Self::Length;
 }
 
+pub trait JavaArrayModificationExt<T: JavaType, RT: JavaScalar>: JvmOp {
+    type SetArrayRegion<'a>: VoidMethod;
+    fn set_array_region<'a>(self, start: usize, values: &'a[RT]) -> Self::SetArrayRegion<'a>;
```
rcoh, line 94: we need to clearly document the invariants that the caller needs to uphold. Is this function always safe to call even if the values are too long, start is out of bounds, etc.? We need tests of that as well

test-crates/duchess-java-tests/tests/java-to-rust/rust-libraries/native_fn_arrays.rs
```diff
@@ -27,3 +29,17 @@ fn break_bytes(num: i64) -> duchess::Result<Java<java::Array<i8>>> {
 
     return Ok(java_array);
 }
+
+#[duchess::java_function(java_to_rust_arrays.JavaArrayTests::fillWithOnes)]
+fn fill_with_ones(arr: Option<&mut duchess::java::Array<i8>>, len: i32) -> i64 {
```
rcoh, line 34: it looks like the API supports having this be `&T`? can you add a ui test that presumably fails if you don't do that?

macro/src/java_function.rs
rcoh: is it possible to get unsoundness through this? If yes, at an absolute minimum we need to make set_array_region unsafe

rcoh (commented):
totally fine with the concept of the API, seems like a good thing to support. We need to really iron out failure modes and edge cases though. I think the fundamental premise of Duchess is that we _fully_ encode everything into the type system so unsoundness becomes impossible.

src/array.rs
bowemi: I think we can ensure safety here. 

I was hesitant to eat the cost of GetArrayLength, but after some testing and some research GetArrayLength seems to just take a few nanoseconds so I might as well call it and guarantee safety then panic if the array length, offset, and slice aren't compatible

test-crates/duchess-java-tests/tests/java-to-rust/rust-libraries/native_fn_arrays.rs
bowemi: I can, but do you think it's worth trying to express mutability given that duchess mostly makes use of ```self``` and not ```&self```? 

In this code, the ```&mut``` is not actually required to mutate the array with ```set_array_region```

src/array.rs
bowemi: Probably the closest thing is ```String```'s ```replace_range```, which uses a range + ```&str```

https://doc.rust-lang.org/std/string/struct.String.html#method.replace_range

Originally I tried to implement ```copy_from_slice```, but ran into some trouble around ```&mut self``` because of only having ```self```.

I could try doing a ```self``` version of ```copy_from_slice```

PR #205, opened 2026-01-06
Support generic exceptions, Rust keywords in Java identifiers

dnsmth:
Fixes two issues around duchess not accepting certain Java code:

1. When the Java identifier has a reserved Rust keyword, use Rust literal syntax. For example, `impl` in Java code should become `r#impl` in the generated Rust. Fixing this required enabling syn's `parsing` to detect Rust keywords, I can find an alternative if this is problematic for some reason.
2. Support throwing generic exceptions, as `throws E` is perfectly legal in Java code.

### DIFF
--- a/duchess-reflect/src/class_info.rs
+++ b/duchess-reflect/src/class_info.rs
@@ -622,7 +622,11 @@ impl Id {
 
     pub fn to_ident(&self, span: Span) -> Ident {
         let data = self.data.replace("$", "__");
-        Ident::new(&data, span)
+        // A legal Java identifier may be an illegal Rust identifier, in which case we'd 
+        // need to use raw syntax. 
+        let mut ident = syn::parse_str(data.as_str()).unwrap_or_else(|_| Ident::new_raw(&data, span));
+        ident.set_span(span);
+        ident
     }
 
     pub fn to_snake_case(&self) -> Self {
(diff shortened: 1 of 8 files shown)


duchess-reflect/src/class_info.rs
```diff
@@ -622,7 +622,11 @@ impl Id {
 
     pub fn to_ident(&self, span: Span) -> Ident {
         let data = self.data.replace("$", "__");
-        Ident::new(&data, span)
+        // A legal Java identifier may be an illegal Rust identifier, in which case we'd 
+        // need to use raw syntax. 
+        let mut ident = syn::parse_str(data.as_str()).unwrap_or_else(|_| Ident::new_raw(&data, span));
```
rcoh, line 627: note that this won't work for a few "ultra reserved words" but seems like a reasonable step

https://github.com/smithy-lang/smithy-rs/blob/main/codegen-core/src/main/kotlin/software/amazon/smithy/rust/codegen/core/rustlang/RustReservedWords.kt#L156-L166

rcoh (approved):

duchess-reflect/src/class_info.rs
dnsmth: Oh, interesting. Maybe I should just test this but is it as simple as doing the same underscore mapping here as is done in the linked code?

rcoh: yeah you just need another non-`r#` binding. I think you might be fine here, I can't remember all the specifics. This is if you have stuff like method parameters. (e.g. you can't have your own parameter named `self` or `Self`)

